PR #1, opened 2016-07-10
Recipes for StatsBase.Histogram

oschulz:


oschulz:
Needs more work - :bar and :heatmap are, in the end, not the dial way to draw histograms.

src/hist.jl
```diff
@@ -1,5 +1,28 @@
 
 # ---------------------------------------------------------------------------
+# StatsBase.Histogram
+
+_hist_edge_centers(edge::AbstractVector) = (edge[2:end] + edge[1:end-1]) / 2
+
+@recipe function f{T, E}(h::Histogram{T, 1, E})
+    x := _hist_edge_centers(h.edges[1])
+    y := h.weights
+    seriestype := :bar
```
tbreloff: I think this should use `-->`, not `:=`.  This way someone can draw a path or something else if they want.

Instead of setting x and y and returning `()`, I'd just return the centers and weights as a tuple.  Otherwise you'll lose proper processing of the fillrange.

oschulz:
> Instead of setting x and y and returning (), I'd just return the centers and weights as a tuple. Otherwise you'll lose proper processing of the fillrange.

Thanks - I was wondering what the recommended way was - I stole the `x := ..., y := ...` from recipes.jl in Plots. ;-)

oschulz:
> I think this should use -->, not :=. This way someone can draw a path or something else if they want.

Well, this will change anyhow, as I'm going away from the simplistic edge-centers to support histograms with non-uniform binning.

tbreloff:
Ok. Just keep in mind... Is this a setting the user might want to override?
Then use '-->'

Regarding setting x/y... The rule of thumb is that "user recipes" and "type
recipes" should return the data arguments, and "plot recipes" and "series
recipes" should return an empy tuple.

oschulz:
Still under development ...

ChrisRackauckas:
Note that the convention is usually to put [WIP] in-front of the PR for "Work In Progress"

oschulz:
Thanks for the tip, @ChrisRackauckas!

oschulz:
I hope to get back on this, soon.

oschulz:
Getting back on this, now.

mkborregaard:
hi @oschulz it would be cool to have some nice work done on histograms (also for Plots itself IMHO)  and this looks like a nice idea. Can you say something about 1) what is missing before this can merge, and 2) what the improvement is with this way of generating the histograms? Thanks!

oschulz:
hi @mkborregaard , this has been dormant for a while due to some other high-priority work. I *do* want to get this merged, but I have to give it another look (a few things have changed in Plots, in the mean time, I think), and polish it up (esp. concerning optional bin boundary display). The main improvements will be speed and clearer display for histograms with many bins, and correct display of histograms with uneven binning.

mkborregaard:
I have a PR on Plots to base histograms generally on StatsBase https://github.com/JuliaPlots/Plots.jl/pull/690
That would make this even more useful (though you may have to rebase this to Plots if it becomes merged).

src/hist.jl
```diff
@@ -1,3 +1,69 @@
...
+            label := ""
+            primary := false
+            ()
+        end
+        markershape := :none
+    end
+
+    x := xpts
+    y := ypts
+    seriestype := :path
```
mkborregaard: why not use :shape?

src/hist.jl
```diff
@@ -1,3 +1,69 @@
...
+    edge, weights = x, y
+
+    axis = d[:subplot][Plots.isvertical(d) ? :xaxis : :yaxis]
+
+    xpts, ypts = _hist_outline_path(edge, weights)
+    if !Plots.isvertical(d)
+        xpts, ypts = ypts, xpts
+    end
+
+    # create a secondary series for the markers
```
mkborregaard: sorry if I'm being slow, but what are the markers for?

src/hist.jl
oschulz: I think `:path` is faster (but I can try `:shape`, for comparison)

src/hist.jl
oschulz: Sometimes, people want markers at the top of the histogram bins. I want to keep this as generic as possible.

src/hist.jl
mkborregaard: cool enough. I just noticed Tom had used Segments and :shape for the boxplot recipe

oschulz:
> @mkborregaard : I have a PR on Plots to base histograms generally on StatsBase

I discussed this possibility with Tom a long time ago. He didn't like adding a dependency like StatsBase (it pulls in quite a few packages) to Plots. That's one of the reasons StatPlots was created.

oschulz:
@mkborregaard : If you're interested, take a look at the discussion at the end of JuliaPlots/Plots.jl#223. I wrote a somewhat lengthy reasoning about how I'd like to handle histograms in the future. (Wanted it done, already, but some other work prevented me from getting serious about it - I hope to rectify that soon, though).

mkborregaard:
Ah. So I remember that discussion on Gitter. I personally like my histograms as boxes. I agree with Tom that the `stephist` could be a nice alternative seriestype for particular cases (e.g. really many data points), but should not replace the histograms we know and love.`stephist` could draw upon the same underlying machinery as a histogram, so the implementation should be very simple. Your current recipe, does that not allow for normal-looking histograms?
I agree with you 100% that histograms must never have spaces between bars (and bar plots always should).
@tbreloff has nodded to adding a dependency on StatsBase to Plots. (StatsBase only requires SpecialFunctions and DataStructures and isn't heavy. Having StatPlots gives the possibility to expand the statistical support way beyond that).

src/hist.jl
```diff
@@ -38,7 +104,17 @@ Plots.@deps cdensity path
...
+    seriestype := :binned
+    (h.edges[1], h.weights)
+end
+
+@recipe function f{T, E}(h::Histogram{T, 2, E})
+    x := _hist_edge_centers(h.edges[1])
+    y := _hist_edge_centers(h.edges[2])
+    z := Surface(h.weights)
+    match_dimensions := true
+    linewidth := 0
```
mkborregaard: This should probably be `linewidth --> 0`, no reason to constrain the user here.

tbreloff:
> > @mkborregaard : I have a PR on Plots to base histograms generally on StatsBase

> I discussed this possibility with Tom a long time ago. He didn't like adding a dependency like StatsBase (it pulls in quite a few packages) to Plots. That's one of the reasons StatPlots was created.

Regardless of what my opinion used to be (this would not be the first time I change my mind), these dependencies are much more intrusive than StatsBase:

```
Distributions
DataFrames 0.0 0.9
KernelDensity
Loess
```

I actually agree with the idea that histograms should be part of StatPlots, but I think it's too disruptive of a change.

Also, the "integer bins don't really work" comment was one of the reasons I built histograms from scratch... I wanted full control.  I really hate the mentality: "sure, you told me you want 24 bins, but according to my algorithm 30 bins will look nicer.  you're welcome." 

FYI I leave this (and similar) PR(s) in your hands.  I don't have the time to properly review/merge.

mkborregaard:
@tbreloff That's fine I will review etc if it goes here on StatPlots, but if it goes to Plots (which I think the StatsBase.Histogram user recipes should, if not the stephist seriestype) it will have to be either you or Patrick, right?

I have to admit that my PR on Plots will edit the bin numbers to look nicer. The user still has full control, of course, if he really wants it - he could just `a = rand(100), histogram(a, bins = linspace(minimum(a), maximum(a), 24)`.

tbreloff:
> it will have to be either you or Patrick, right?

@mkborregaard I think you can merge PRs in Plots... if not I will add you.

> I have to admit that my PR on Plots will edit the bin numbers to look nicer. The user still has full control, of course, if he really wants it

Yeah I suppose that's true.  I leave the final decision to you.

oschulz:
> @mkborregaard: I personally like my histograms as boxes.

My argument in the original discussion (and I think it still holds) was that steps should be the default, because they *always* look Ok. Seperation lines between bins (a.k.a. boxes) only look Ok in some cases (few bins). But if a majority of users want lines between the bins by default, I could live with that (as long as they're easy to disable).

> Your current recipe, does that not allow for normal-looking histograms

Well, "normal" is subjective here, I think. ;-)

Not yet, but "box-style" (e.g. lines between bins) was one of the things I wanted to add. Ideally, these lines can even look differently than the "step-line" around the hist (e.g. thinner / weaker). I think this could look really neat. Let me get back on it and get this PR in shape quickly  - I really left this stagnant for waaay too long.

mkborregaard:
Sounds good. I'd be reluctant to change the defaults, but I think it is nice to add the possibility.

mkborregaard:
If this should go into Plots it shouldn't add anything that requires extra keywords, but you could experiment with letting lines between boxes be regulated by `markerstrokewidth` and the outside line by `linewidth`, that would give fine-grained control without getting in the way of something else (I think).

oschulz:
> shouldn't add anything that requires extra keywords, but you could experiment with ...

Thanks - I was indeed wondering what keywords to use, since I don't want to introduce additional ones, if possible.

mkborregaard:
@oschulz I have been looking at the code, but I don't see why a lot of the code is necessary. To get the stephist, can't you just do:
```julia
@recipe function f{T, E}(h::Histogram{T, 1, E})
    seriestype := :steppre
    ylims --> [0, maximum(h.weights)*1.1]
    h.edges[1], [0; h.weights]
end
```
?
That seems to do what you want, e.g. if we add a recipe for vectors of Histograms you could get:
```julia
@recipe function f{H <: StatsBase.Histogram}(hv::Vector{H})
    for h in hv
        @series begin
            h
        end
    end
end

plot([fit(Histogram, randn(10000), nbins = 100) for x in 1:3])
```
skaermbillede 2017-02-23 kl 09 51 53

oschulz:
> To get the stephist, can't you just do

I can't use steppre, because it doesn't return to zero:

plot

Also, I definitely think we the recipe for histogram should be defined via the new series type `:binned`, so that people can use that to plots histogram-like data that is not a StatsBase histogram.

Give me a little time to get this in shape (it's really just a draft), and then we can review? I would definitely some input, but I got to get this in a decent state, first. :-)

tbreloff:
> I can't use steppre, because it doesn't return to zero

Of course you can just add a (1,0) point...

> I definitely think the recipe for histogram should be defined via the new series type :binned

I admit I don't remember everything we discussed on this topic, but every effort should be made to keep this simple.  If adding a `:binned` recipe will help simplify things, then great.

tbreloff:
> it'll result in a zero-length horizontal line at the end

I don't think this is a problem.

> More importantly though, steppre puts the markers at the points, not in the bin centers, which makes no sense for a histogram.

This is valid.  I'm not sure how often people put markers, but I could see adding error bars.

PGFPlots supports a `:stepmid` type... if that was implemented as a recipe it would get part of the way.  But maybe the `:binned` type is the rest of the way?  I don't remember well enough what that is exactly

oschulz:
> I'm not sure how often people put markers, but I could see adding error bars.

Yep, forgot to mention errors bars - already working on them.

oschulz:
I hit a problem while implementing error bars - I can't find a way to manipulate error bars in a user recipe. Example:

```julia
@recipe function f(::Type{Val{:some_recipe}}, x, y, z)
    yerror := nothing
    seriestype := :path
    ()
end

plot(1:10, rand(10), yerror = 0.2, grid = false, seriestype = :some_recipe)
```

This results in a plot *with* error bars, in spite of the `yerror := nothing`. Is there a way around this? Obviously, I need to disable the default error bars, and add custom ones at the bin centers (e.g. together with the markers).

mkborregaard:
Error bars on a histogram? What is the use case here? The only one I can think of is combining several histograms to look at the variation. This kind of functionality is implemented by @piever in the GroupedError recipe of StatPlots, for bars and density plots - it may work for histograms out-of-the box or need very little tweaking to do this, so that is worth checking out.

I am sorry for disagreeing here, but I think a simple step(pre/post/mid) solution is best. This could be implemented as a :stephist recipe that calls the same functionality as :histogram currently except that it specifies step(post/pre). Adding a shorthand would give you the default behaviour you are requesting (simply call `stephist(x)`) without changing the look of histogram.

A :stepmid series type would be of general benefit to Plots, and should be simple to implement based on steppost and steppr. That would allow you to add error bars manually to the bin centers Again, I would like to see  a use case that cannot be better achieved with groupapply. The only issue with stepmid is that it would likely cause a half-bar-width line at y = 0 on each side of the histogram, depending on how we would like stepmid to behave. The functionality could be applied to the :bar histograms by calling `centers` and defining bar_width to force the bars to touch.

Here's how I would suggest splitting this PR:

I think the two basic recipes for 1-d (with seriestype := :bar at the moment) and 2-d StatsBase.Histogram belong in Plots and could be merged there now if you'd like to make a PR? I also think a stepmid (importantly, the code should be as similar to steppost as possible) series recipe could go right into there. I could then add the :stephist recipe once this (JuliaPlots/Plots.jl#690) is merged. This would allow everything to work with Plot's inbuilt error bars, and be minimally disruptive (it adds 1 or 2 new series recipes that are consistent with the existing).

If you have some more complex ideas for custom markers etc. I would think this would belong better as a special recipe in PlotRecipes. That would give you freedom to be creative with the histogram and create exactly the look you prefer.

piever:
> Error bars on a histogram? What is the use case here? The only one I can think of is combining histograms for different groups to look at the variation. This kind of functionality is implemented by @piever in the GroupedError recipe of StatPlots, for bars and density plots - it may work for histograms out-of-the box or need very little tweaking to do this, so that is worth checking out.

Error bars for density plots are already implemented. To have the non-normalized version, you would simply need to add an option [here](https://github.com/JuliaPlots/StatPlots.jl/blob/master/src/groupederror.jl#L58) to not renormalize the count.
However there is no binning implemented, it only works for discrete data where the binning is provided by the set of values taken by the data. So the easiest to get error bars on histogram is to add a "binning" method that takes a column of a dataset `df[:x]` and adds to the dataframe the column `df[:x_binned]`, where each element represents the identity of the bin of the corresponding value of x. Then the `groupapply` machinery could be applied to the binned column, giving binned density plot/histograms with errorbars and a bunch of other functionality.

mkborregaard:
Why not let StatsBase.histogram do the binning, like you do with the density plot?

mkborregaard:
I am not saying we necessarily need this, I just want to make sure operations are consistent and existing functionality leveraged where possible.

piever:
I see. I think there are two not necessarily exclusive option:

- Add a new analysis function (just like` _density`, `_cumulative` and `_locreg`) whose purpose is to count data in preexisting bins. This could be from StatsBase. It's probably improtant to notice that once you have the bins, you don't really need StatsBase: they basically run `searchsortedfirst` on the edges vector for every datapoint

- More generally, if we keep not only the bin count, but also what datapoint belongs to what bin (which comes for free once you run `searchsortedfirst`) you get density and histogram, but also all the analysis that one could implement for discrete data (e.g. average value of y for a given value of x, or more sophisticated stuff that is not yet implemented) is now also available for continuous data.

So far my preference is for option 2.

oschulz:
> Error bars on a histogram? What is the use case here?

It may depend on the discipline, but in particle physics, for example, it's very common to have error bars on a histogram to indicate the statistical error.

oschulz:
> @mkborregaard: I am sorry for disagreeing here, but I think a simple step(pre/post/mid) solution is best.

I'm sorry, but neiter of these will give you correct bins for uneven binning - and neither will bar. step-xxx takes n inputs for the weights and n inputs for the positions - that doesn't work for a histogram, as there are n+1 bin boundaries. The current bar-based histogram recipe fails miserably for uneven binning, too.

A binned plot should be a stepped path, from y=0 up to the value of the first bin, horizontal over each bin, down to y=0 again after the last bin, taking the individual bin widths into account. This is what I'm implementing with `:binned`. Whether there are lines between the bins is a matter of preference, and this should be supported, but it must be optional (doesn't work with many bins when overlaying several hists).

I'm all for re-using existing functionality - but I'm convinced that there is a fundamental difference between step-xxx, bar and binned. Of course I'll try to re-use what we have in `:binned`, under the hood, but some things are just different there.

And error bars are *very* common. In fact, another draw style I want to support for `:binned` all lines off, and bins represented by horizontal and vertical error bars - these plots are quite common in my field, too.

oschulz:
> So the easiest to get error bars on histogram is to add a "binning" method that takes a column of a dataset df[:x] and adds to the dataframe the column df[:x_binned]

Uh, introducing a `Dataframe` to merely plot pre-binned data and add error bars doesn't strike me as very simple ... :-)

oschulz:
> @mkborregaard: :stepmid series type would be of general benefit to Plots, and should be simple to implement based on steppost and steppr. That would allow you to add error bars manually to the bin centers

But how do I add the error bars manually? Like I wrote, I can't find a way to modify `xerror` and `yerror` in a recipe. I just need to know how to do that, then the rest will be simple, and I can get this cleaned up and ready to merge (for 1D binned) very quickly.

oschulz:
> If you have some more complex ideas for custom markers etc. I would think this would belong better as a special recipe in PlotRecipes (or it might still belong here in StatPlots).

I think a user should be able to do `plot(hist, ...)` and get common things like proper error bars (and maybe even markers) out of the box. I do agree that this should be in Plots (and the `:binned` recipe, too).

tbreloff:
I'm pretty sure "error" is the attribute, and xerr/yerr are conveniences.
But I'd have to look at the code to remember exactly.

As for stepmid vs others... i would agree that "binned" or something should
be independent with a primary filled step series and a secondary scatter
series at the middle of each bar.

It shouldn't be any more complex than that, and more complex histogram
functions should just create a "binned" series.

One last point (sorry not in the right issues) is that none of this should
be specific to DataFrames. If it is, you're doing it wrong and going
against the core essence of Plots. Please make sure this could be used with
any reasonable input data.

Thanks for all the effort.

mkborregaard:
@oschulz > neither of these will give you correct bins for uneven binning - and neither will bar

Good point. OK consider me convinced :-) 

@tbreloff > "error" is the attribute, and xerr/yerr are conveniences.

No the attribute is `yerror` - check `plotattr(:Series)` :-) 
I think there must be a bug here. Supplying `yerror := []` doesn't help.

@oschulz > none of this should be specific to DataFrames

of course.

mkborregaard:
But let me just understand why you want to override the inbuilt error bars?

src/hist.jl
tbreloff: I think `:path` is preferable here

tbreloff:
I just looked at the code, and it seems to be the method I would have used as well... great.  Assuming you've tested this, and it doesn't break existing functionality, I say we merge it.

oschulz:
> But let me just understand why you want to override the inbuilt error bars?

I don't want to override the style of the error bars - I just need them to show up at the bin centers. Like Tom wrote, the step-path shouldn't have any error bars, and there should be a secondary scatter plot at the bin centers for markers and/or error bars. But no matter what I try, the error bars always show up at the original edge/weight coordinates that go into `:binned`. I have a feeling that somehow the Plots pipeline handles `:xerror` and `:yerror` specially.

oschulz:
> I just looked at the code, and it seems to be the method I would have used as well... great. Assuming you've tested this, and it doesn't break existing functionality, I say we merge it.

Uh, which code, now? :-) Do you mean this PR?

tbreloff:
> Do you mean this PR?

I did.  But if you don't think it's ready then that's fine too.

> But no matter what I try, the error bars always show up at the original edge/weight coordinates that go into :binned. I have a feeling that somehow the Plots pipeline handles :xerror and :yerror specially.

Yes error bars are handled in a special way.  And frankly I'm not sure of the right way to avoid this behavior.  Right now error bars are extracted into their own series just after user recipe processing.  Search for `:xerror` in the `src/pipeline.jl` file if you want to trace through what happens.  It might be possible to push this handling to later in the pipeline, but I don't have the capacity right now to work on that.

I think you should just ignore error bars and get the `:binned`/`:stephist` functionality to a good stopping point.

oschulz:
> I did. But if you don't think it's ready then that's fine too.

Thanks - I'd prefer to finish it (I promise, for real, now, and quickly!) instead of merging it half-baked and starting a new one immediately.

The question is what to do about JuliaPlots/Plots.jl#690? If that's to be merged, it may be better to do that first and close this PR. Then I'd open up a new one in Plots and implement `:binned` there and modify the plotting recipe for `Histogram` to use it. Or I could move this PR over to Plots now, on top of JuliaPlots/Plots.jl#690, and we'd cancel 690 and merge the new one once it's done.

What would you guys prefer?

tbreloff:
Assuming `histogram(y)` is going to return a "bar plot" as it does now, and `stephist(y)` will return a `:binned`, then it doesn't matter where this PR goes.  But I'd be perfectly fine to add this in Plots next to `:histogram`.

mkborregaard:
I'd personally prefer to merge this to Plots when it works. I'll wait and rebase 690 based on this code to make all histograms use this functionality.
With regards to lines between bars, could the current PR lead to 1) binned, that looks like the current histogram, and 2) stephist, that only has a path, no fill? That's what I'd prefer.

oschulz:
> I'd personally prefer to merge this to Plots when it works. I'll wait and rebase 690 based on this code to make all histograms use this functionality.

Ok, then I'll see that I get this in shape as quickly as possible.

> With regards to lines between bars, could the current PR lead to 1) binned, that looks like the current histogram, and 2) stephist, that only has a path, no fill?

No, I'd really like :binned to support both fill, optional bin-separation lines, and optional error-bars (which can look different than the line on top). This is still qualitatively different from bars (see the discussion in JuliaPlots/Plots.jl#223).

The plotting recipe for Histogram should, in my opinion, use then use `:binned` recipe by default - maybe with the option for the user to select `:bar` style. As a quick recap of the discussion in JuliaPlots/Plots.jl#223: A histogram is, fundamentally, a visualization of categorized data. Depending on whether the categories are discrete or represent numerically contiguous bins, bars (arbitrary width, may have gaps between them) or bins (no gaps, single or no line between) are the natural drawing style. `StatsBase.Histogram` clearly falls into the category "numerically contiguous bins".

tbreloff:
It just feels like you're trying to cram too much in one recipe. Recipes
can pass to other recipes... lets get something basic working and merged,
and expand it later. I think the reason your PRs have sat around unmerged
is partially because you try to do too much all at once.

mkborregaard:
A histogram is always based on binning continuous data. If it is categories, it is not a histogram, but a bar plot. I think that this generally is consistent with what you are saying.

I'd suggest you create two series recipes - it doesn't matter what you call them, but I think the user-facing names should be :histogram that looks like this:
skaermbillede 2017-02-24 kl 17 50 30
and :stephist. Both of the recipes should use the hist_outline function to create the path vertices. But the stepped one should create a :path (:binned), the default one used by histogram should draw the bars one by one as Shapes.

mkborregaard:
But as Tom says let's start with merging your binned type, which you already have. The rest can follow. :-)

oschulz:
> A histogram is always based on binning continuous data.

Not necessarily - if your data stems from a discrete statistical distribution (e.g. Poisson), the values are integers, and there is no bin width. Still, most people would call the resulting plot a histogram (at least I don't know what else to call it) - and in this case, "bar-style" (bar-width is non-informative) or a scatter-plot (e.g. when drawing a fit too, to keep things uncluttered) is the "natural" plotting style (unless the number of values is high, so that bars can't be plotted properly anymore).

I'd suggest to cleanly separate between semantics (I want to plot a histogram) and visualization (bar, binned or scatter-plot). So when plotting a histogram using `plot(myhist)` the user should be able to select the drawing style. But as, indeed, usually a `StatsBase.Histogram` represents contiguous bins, I think `:binned` should be the default.

> I think the user-facing names should be :histogram

For the reasons above (a histogram is, in my opinion, a semantic concept, not a drawing style), it'd prefer to call the series recipes `:bar` and `:binned` or similar.

> But as Tom says let's start with merging your binned type, which you already have. The rest can follow. :-)

So be it. If you let me improve on it, afterwards. :-)

oschulz:
Another thing I forgot: Calling the recipe :histogram might imply that it takes unbinned-data and bins them, to some people. Maybe I should call my series recipe :bins or :bin1d, instead of :binned, to avoid confusion?

mkborregaard:
It doesn't matter what you call it - I'd suggest adding a `d[:seriestype] == :stephist && seriestype := :binned` to the StatsBase.Histogram user recipe.

oschulz:
Well, maybe it's really better to get this merged, and then worry about error bars, etc. Let me make a few little changes, test it one more time, and then I'll remove the [WIP] label tonight.

mkborregaard:
And transfer to Plots?

oschulz:
Ok, just rebased the PR, this should be Ok to merge now. Plotting works for both 1D (e.g. via `:bins`) and 2D (via `:bins2d`) histograms - here's an example:

hists-demo

For 1D, series types `:bins` (default), `:scatter` (xerror automatically set to bin-width), and `:bar` (not shown in the example) are supported. Bar plots don't work for inhomogeneous binning, currently (has to be fixed in Plots).

One thing we should get on soon though: log-scale (y-axis) doesn't work - but that seems due to issues in Plots. log-scale histogram plots are *very* important in many disciplines (common for spectra, for example).

The `:bins2d` series recipe is basically a slightly modified version of the `:histogram` recipe in Plots - with the binning removed, as it's for data that is already binned. If this moves into Plots, then `:histogram` can be simplified and use `:bins2d`.

The question is - should we merge all of this into StatPlots initially, or add at least `:bins` and  `:bins2d` to Plots directly, or close this PR and merge it all into Plots at once?

tbreloff:
These look great.

Can you confirm... a histogram would pass to a bins, and a histogram2d
would pass to a bins2d?  What distinguishes a bins2d from a heatmap?

As for moving to Plots. Lets do that now. If there are extensions for
complex alternatives to standard histograms then we can add those
extensions to StatPlots later.

Thanks for your effort.

oschulz:
> Can you confirm... a histogram would pass to a bins, and a histogram2d would pass to a bins2d?

Yep.

> What distinguishes a bins2d from a heatmap

It uses heatmap, but it add's the bin-centers for x and y and sets zero-valued bins to NaN. It's basically just the Plots `:histogram2d` recipe (that's where I stole the code from) with the binning code taken out. So once `:bins2d` is in Plots, `:histogram2d` can just use it (basically a code transfer).

oschulz:
> As for moving to Plots. Lets do that now.

Just `:bins` and `:bins2d`, or also the recipes for `StatsBase.Histogram` (making StatsBase a requirement for Plots)?

tbreloff:
This is where you combine with the other PR, right? I'll let you two sort
it out ;)

mkborregaard:
Move all of it :-)
I have one suggested change - make :bar the default, not :bins. I strongly feel that 1) plotting a StatsBase histogram should look the same as calling `histogram` on an array, 2) that we should not change the default look of `histogram` - which would be a major breaking change away from a commonly accepted norm.

oschulz:
> Move all of it :-)

Then I'll create a new PR for Plots, and change this one to simply remove the old Histogram recipe from StatPlots - Ok?

> Pretty plots! (could you post the calling code for those?)

I's a bit involved, because the hists with inhomogeneous binning are normalized. I have some code that implements `Base.normalize!` for Histogram that I want to contribute to StatsBase, but I haven't gotten around to do a PR yet (will need to write testing code, first, otherwise it won't get accepted).

mkborregaard:
It'd just be nice to see what you have in mind, especially if I modify later. The normalization should affect nothing but the y axis even with uneven bins, right? 
Could you do the PR on this new branch: https://github.com/JuliaPlots/Plots.jl/tree/new-histogram
I'll add mine after, and we can talk together before merging the whole thing into master.

mkborregaard:
Note that uneven bins for bars on the existing master could be fixed with this 2-word PR: https://github.com/JuliaPlots/Plots.jl/pull/705 (though there is a bug with the heights).

oschulz:
> I have one suggested change - make :bar the default, not :bins. [...] that we should not change the default look of histogram

Well, I guess people have gotten used to it, so, Ok. :-)

> normalization should affect nothing but the y axis even with uneven bins, right

Right, that would go in StatsBase, though, not in Plots. I'll try to make that PR soon, too. After I get that into StatsBase, we can simplify the current normalization code in Plots by using it.

> Could you do the PR on this new branch: https://github.com/JuliaPlots/Plots.jl/tree/new-histogram

Sure, will do - good idea!

> Note that uneven bins for bars on the existing master 

You beat me to it. :-) Suggestion: Merge this into the new-histogram branch, then we'll have it all together.

mkborregaard:
> Merge this into the new-histogram branch

https://github.com/JuliaPlots/Plots.jl/pull/707 Also fixed the heights bug.
Looking at this made me realize that there is the same bug in bin heights for uneven bins in StatsBase too, and I've opened an issue: https://github.com/JuliaStats/StatsBase.jl/issues/241

> I'll try to make that PR soon, too

You should make it immediately, and disregard that you don't have the testing code for now. Someone there is likely to be working on it. You should have an easy time getting this in, as Edelman himself requested this last Friday: https://github.com/JuliaStats/StatsBase.jl/issues/231

> It's a bit involved

I'd just like to see the `plot` and `plot!` calls.

oschulz:
> You should have an easy time getting this in, as Edelman himself requested this last Friday: JuliaStats/StatsBase.jl#231

Here it is: JuliaStats/StatsBase.jl#243.

> I'd just like to see the plot and plot! calls.

Sure (slightly changed by now):

```julia
data = filter(x -> abs(x) < 2.5, randn(10000))

h1 = fit(Histogram, data, -4:0.1:4, closed = :left)
h1_norm = normalize(h1, mode = :pdf)

inhom_edge = map(x -> x + 0.1 * x^3 / 40, -4:0.5:4)
h2 = fit(Histogram, data, inhom_edge, closed = :left)
h2_norm = normalize(h2, mode = :pdf)

h2d = fit(Histogram, (randn(1000000), 0.1 * rand(1000000)), nbins = (40, 50), closed = :left)

plot(
    plot(
        h2_norm,
        grid = false,
        fillrange = 0,
        fillalpha = 0.5,
        fillcolor = :red
    ),
    plot(
        h2_norm,
        grid = false,
        yerror = sqrt.(h2.weights) .* h2_norm.weights ./ h2.weights,
        seriestype = :scatter
    ),
    plot(
        [h1_norm, h2_norm],
        xlims = (-3, 3),
        grid = false,
    ),
    plot(h2d),
    grid = false,
    size = (1024, 768)
)ww
```

hists-demo

oschulz:
Got the PR for Plots.jl/new-histogram almost ready - few minor kinks, hope I can get it finished today or tomorrow. I think you guys will like it.

oschulz:
Changed to remove StatsBase support from StatPlots, functionality is moving to Plots: JuliaPlots/Plots.jl#713.

oschulz:
We may want to squash this down (to just removing StatsBase support, without the intermediate development) before merging.

mkborregaard:
> We may want to squash this down (to just removing StatsBase support, without the intermediate development) before merging.
Not sure what you mean? I will merge this when we merge the Plots PR. Thanks!

mkborregaard:
I'll just close this.

oschulz:
Uh, but don't we need to merge this? Otherwise, we'll have conflicting recipes for StatsBase.Histogram in StatPlots and Plots.

mkborregaard:
Absolutely. I just thought that the easiest approach was for one of us to open a new PR just removing the StatsBase recipe, and hold that open until the Plots one gets merged. Happy to reopen if you disagree.

oschulz:
Oh, sure - doesn't matter. But then I would suggest to open that PR already, so we don't forget about it. :-)

mkborregaard:
good idea - would you?

mkborregaard:
I don't mind either, it's just that you'd get the credit.

oschulz:
> I don't mind either, it's just that you'd get the credit.

Lol, the credit will be in the commit to Plots, I guess. :-)

Anyhow, I just modified this PR to do what's necessary - can you just reopen it? Then we don't need a new one. It'll give me the satisfaction to see this merged eventually, after such a long time. :-)

mkborregaard:
Sorry, can't reopen because the branch was force-pushed :-(

oschulz:
Arg, GitHub! :-)

Ok, replaced by #43.

PR #5, opened 2016-07-16
Update hist.jl

odow:
`isvertical` is not exported from `Plots.jl`


tbreloff:
I'm very sorry that I never merged this.  And instead I merged someone else's that did the same thing!  Hopefully you'll get on the contributors list another way!

odow:
No worries @tbreloff! I'd totally forgotten about this

PR #11, opened 2016-09-27
Add cumulative density recipe

anriseth:
I often look at the cumulative density of data sets in addition to the approximate pdfs.
Putting a PR here in case it is useful for anybody else.

(I put the code in hist.jl because the density recipe was there)


tbreloff:
Thanks!

PR #12, opened 2016-09-28
Let smooth be optional for cornerplot

mauro3:
This is probably not how it is done properly, but it works.  Trying out the dirty-hack strategy again ;-)


tbreloff:
Can't you just change ':=' to '-->'?

mauro3:
I tried a few iterations of that but it didn't seem to work. you mean on line https://github.com/JuliaPlots/StatPlots.jl/blob/master/src/cornerplot.jl#L100 , right?

tbreloff:
Right

tbreloff:
f34ff30

Issue #24, opened 2016-11-29
Cannot plot multiple series from DataFrame


mkborregaard:
I cannot plot multiple series from a DataFrame with StatPlots. I can do it with the old Plots on my julia-0.4.6 install.

so:
```julia
using StatPlots
df = DataFrame(a = 1:100, b = randn(100), c = randn(100))
plot(df, :a, :b)
```
works, but
```julia
plot(df, :a, [:b :c])
```
doesn't, but it should, since
```julia
a = 1:100 ; b = randn(100); c = randn(100)
plot(a, [b, c])
```
works.

mkborregaard:
Error message:
```
LoadError: MethodError: Cannot `convert` an object of type Array{Float64,1} to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
while loading In[81], in expression starting on line 1

 in discrete_value!(::Plots.Axis, ::Array{Any,1}) at /Users/michael/.julia/v0.5/Plots/src/axes.jl:428
 in expand_extrema!(::Plots.Subplot{Plots.PlotlyJSBackend}, ::Dict{Symbol,Any}) at /Users/michael/.julia/v0.5/Plots/src/axes.jl:294
 in _expand_subplot_extrema(::Plots.Subplot{Plots.PlotlyJSBackend}, ::Dict{Symbol,Any}, ::Symbol) at /Users/michael/.julia/v0.5/Plots/src/pipeline.jl:361
 in _process_seriesrecipe(::Plots.Plot{Plots.PlotlyJSBackend}, ::Dict{Symbol,Any}) at /Users/michael/.julia/v0.5/Plots/src/pipeline.jl:392
 in _plot!(::Plots.Plot{Plots.PlotlyJSBackend}, ::Dict{Symbol,Any}, ::Tuple{DataFrames.DataFrame,Symbol,Array{Symbol,2}}) at /Users/michael/.julia/v0.5/Plots/src/plot.jl:229
 in #plot#258(::Array{Any,1}, ::Function, ::DataFrames.DataFrame, ::Vararg{Any,N}) at /Users/michael/.julia/v0.5/Plots/src/plot.jl:52
 in plot(::DataFrames.DataFrame, ::Symbol, ::Vararg{Any,N}) at /Users/michael/.julia/v0.5/Plots/src/plot.jl:46
```

mkborregaard:
[...]

PR #25, opened 2016-11-29
allows the passing of expressions including DataFrame columns as symbols

mkborregaard:


src/dataframes.jl
```diff
@@ -21,12 +21,31 @@ function Plots.extractGroupArgs(group::Symbol, df::AbstractDataFrame, args...)
     Plots.extractGroupArgs(collect(df[group]))
 end
 
+# allows the passing of expressions including DataFrame columns as symbols
+function processExpr(expr, df)
+    arg = map(expr.args) do x
+        isa(x, Expr) && return processExpr(x, df)
+        if isa(x, QuoteNode) && isa(x.value, Symbol)
+            return collect(df[x.value])
```
tbreloff: I don't think this is right... you don't want to collect here... you want to return a **collect expression** so that it will be called during the `eval`.  (it may work, but that requires building a giant constant expression to be evaluated)

mkborregaard: Yes it does seem to work. But I will look into doing it the proper way as you suggest.

tbreloff: Hopefully this example clarifies what you're doing:

```julia
julia> ex = :(:x + 1)
:(:x + 1)

julia> dump(ex)
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol +
    2: QuoteNode
      value: Symbol x
    3: Int64 1
  typ: Any

julia> ex.args[2] = rand(10)
10-element Array{Float64,1}:
 0.573531 
 0.0145821
 0.531146 
 0.683965 
 0.439694 
 0.24342  
 0.314299 
 0.540411 
 0.996641 
 0.346173 

julia> dump(ex)
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol +
    2: Array{Float64}((10,)) [0.573531,0.0145821,0.531146,0.683965,0.439694,0.24342,0.314299,0.540411,0.996641,0.346173]
    3: Int64 1
  typ: Any

julia> eval(ex)
10-element Array{Float64,1}:
 1.57353
 1.01458
 1.53115
 1.68397
 1.43969
 1.24342
 1.3143 
 1.54041
 1.99664
 1.34617
```

It works, but you're overworking `eval`

mkborregaard: Yes, that is exactly the operations I was going through when developing this :-) I just didn't realize it was wrong.
So you suggest instead of passing in collect(df[x.value]) I should pass :collect(df[:($x.value)])?

mkborregaard: I have tried moving `eval` out as the last step but still doesn't work. I will keep looking at this
```julia
# allows the passing of expressions including DataFrame columns as symbols
function processExpr!(expr, df)
    arg = map(expr.args) do x
        isa(x, Expr) && processExpr!(x, df)
        if isa(x, QuoteNode) && isa(x.value, Symbol)
            return :(collect(df[:($(x.value))]))
        end
        x
    end
    expr.args = arg
end

# if a DataFrame is the first arg, lets swap symbols out for columns
@recipe function f(df::AbstractDataFrame, args...)
    # if any of these attributes are symbols, swap out for the df column
    for k in (:fillrange, :line_z, :marker_z, :markersize, :ribbon, :weights, :xerror, :yerror)
        if haskey(d, k)
            if isa(d[k], Expr)
                processExpr!(d[k], df)
                d[k] = eval(d[k])
            end
```

tbreloff: I think just: `return :(collect(df[$x]))`

Also you are right about recursion... you can just add an "elseif" clause to the QuoteNode check if `arg` is an `Expr`, and `return processExpr!(x, df)`

mkborregaard:
I think this should work - I had to interpolate `df` as well, which took me some time to realize...

tbreloff:
Thanks for the effort @mkborregaard.  I merged and any changes/additions can be worked out in a new PR/commit.

mkborregaard:
Wohoo! Is there also an obvious place to look to change so I can do stuff like `plot(df, :a, [:b :c])`?

tbreloff:
Well that one should work already.  The dataframes.jl file is pretty small... shouldn't be too hard to narrow down.

mkborregaard:
OK, I will have a look (it didn't work yesterday #24)

PR #27, opened 2016-11-30
Added shaded error plots

piever:
I've added the shaded error plot (which was trivial to do using ribbon, except ribbon plots are not implemented in all backends, a notable exception being PlotlyJS: [#565](https://github.com/tbreloff/Plots.jl/issues/565) ).

screenshot 2016-11-30 20 08 43

Also, I thought it'd be nicer to be able to plot shaded error bars starting directly from a dataframe, giving as input also the variable corresponding to the x axis, the function corresponding to the analysis of interest (could be a density plot, or a cumulative plot, or plotting some other column y as a fct of x) and the categorical variable corresponding to the "population of interest" across which one should compute the standard error of the mean to see to what extent the analysis is consistent across individuals of the population (this kind of analysis is common in my field). Also, one can split data according to one or more other columns.

Example: compute density plot of magnitude of earthquakes. Shaded-error bar = s.e.m. across stations. Split data by depth of earthquake.

screenshot 2016-11-30 20 08 52

This is preliminary work, but if it is of interest I think it could be nice to continue in this direction as follows:

- In the version where the data is conditioned on one or several variables, add possibility of modifying graphical elements according to the value of those variables (label, color, dashed vs full line and other attributes could depend on the value of the conditional variables in a way determined by the user)
- Build a list of reasonable functions f and incorporate them, e.g. density, cdf, hazard, y as function of x (with some nonparametric regression I guess)
- Do the same for scatter plots and bar plots

### DIFF
--- a/src/StatPlots.jl
+++ b/src/StatPlots.jl
@@ -23,5 +23,6 @@ include("violin.jl")
 include("hist.jl")
 include("marginalhist.jl")
 include("bar.jl")
+include("shadederror.jl")
 
 end # module
--- a/src/shadederror.jl
+++ b/src/shadederror.jl
@@ -1,0 +1,106 @@
+function border(x,y,rib)
+    rib1, rib2 = if Plots.istuple(rib)
+        first(rib), last(rib)
+    else
+        rib, rib
+    end
+    yline = vcat(y-rib1,(y+rib2)[end:-1:1])
+    xline = vcat(x,x[end:-1:1])
+    return xline, yline
+end
+
+@recipe function f(::Type{Val{:shadederror}},plt::Plot; shade = 0.)
+
+    # set up the subplots
+    x,y = d[:x],d[:y]
+    xline, yline = border(x,y,shade)
+
+    # line plot
+    @series begin
+        primary := true
+        x := x
+        y := y
+        seriestype := :path
+        ()
+    end
+
+    @series begin
+        # shaded error bar
+        primary := false
+        fillrange := 0
+        fillalpha --> 0.5
+        linewidth := 0
+        x := xline
+        y := yline
+        seriestype := :path
+        ()
+    end
+end
+
+@shorthands shadederror
+
+## Define a more general shadederror function that can take as input:
+## - a function to be applied to your dataframe
+## - a dataframe
+## - the x-variable of the plot
+## - the categorical variable used to compute s.e.m. across population
+## - optionally, another categorical variable used to split the data
+
+get_axis(column::PooledDataArray) = sort!(union(column))
+get_axis(column::AbstractArray) = linspace(minimum(column),maximum(column),100)
+
+# f is the function used to analyze dataset: define it as NA when it is not defined,
+# the input is: dataframe used, x variable and points chosen on the x axis
+# the output is the y value for the given xvalues
+
+function get_mean_sem(f, df, x, population)
+    # define points on x axis
+    xvalues = get_axis(df[x])
+
+    # get mean value and sem of function of interest
+    splitdata = groupby(df, population)
+    v = DataArray(Float64, length(xvalues), length(splitdata));
+    for i in 1:length(splitdata)
+        v[:,i] = f(splitdata[i],x, xvalues)
+    end
+
+    mean_across_pop = DataArray(Float64, length(xvalues));
+    sem_across_pop = DataArray(Float64, length(xvalues));
+    valid = Array(Bool, length(xvalues));
+    for j in 1:length(xvalues)
+        mean_across_pop[j] = mean(dropna(v[j,:]))
+        sem_across_pop[j] = sem(dropna(v[j,:]))
+        valid[j] = (length(dropna(v[j,:]))>1)
+    end
+
+    return xvalues[valid], mean_across_pop[valid], sem_across_pop[valid]
+end
+
+function shadederror(f, df, x, population; kwargs...)
+    x,y, shade = get_mean_sem(f, df, x, population)
+    shadederror(x,y; shade = shade, kwargs...)
+end
+
+function shadederror!(f, df, x, population; kwargs...)
+    x,y, shade = get_mean_sem(f, df, x, population)
+    shadederror!(x,y; shade = shade, kwargs...)
+    return
+end
+
+function shadederror(f, df, x, population, conditioned; kwargs...)
+    s = plot()
+    shadederror!(f, df, x, population, conditioned; kwargs...)
+    return s
+end
+
+function shadederror!(f, df, x, population, conditioned; kwargs...)
+    if isempty(conditioned)
+        shadederror!(f, df, x, population; kwargs...)
+    else
+        by(df,conditioned) do dd
+            shadederror!(f,dd,x,population; kwargs...)
+            return
+        end
+    end
+    return
+end


tbreloff:
At first glance this looks nice, but it will take me a bit to do a complete review.  Can anyone else review and comment? Thanks for the effort... looking forward to incorporating it.

piever:
In general it seems to me that adding error bar depending on variability across the population in a plot is one of the rare examples of a common plot pattern that is not a "one-liner" in Plots.jl
I'm open to discussion as to what is the best implementation (it doesn't need to be the one I'm proposing here): in particular the syntax choice is I guess open to discussion and in the case of conditioning on some variables and plotting in different colours I'm using a primitive solution that doesn't really take advantage of Plots.jl and could definitely be improved.
Still I think this "errorbar" feature should be added  somehow. Also, it's very easy to extend this pattern to scatter plots and bar plots using metaprogramming. If you think that's valuable I'm happy to add that as well.

tbreloff:
merging and we can fix up later if needed... thanks!!

mkborregaard:
This looks really cool, though I am not sure I understand the example :-) I think the shaders are really nice - they are relevant for this issue: https://github.com/tbreloff/Plots.jl/issues/340

One thing I would love StatPlots to have is recipes for statistical model - e.g. a categorical anova might plot automatically as a bar plot with error bars and asterisks for significance; in this case a linear model with a single predictor could be nice to plot as a line with these shaded errors as a default. Do you think that should be relatively easy to implement with this, @piever ?

piever:
I guess it'd be useful to see an exact example of what syntax you'd want to do what thing. If the concern is bar plots, the same as I've done with line plots and shaded errorbars can be done with scatter plots (or bar plots) and normal errorbars, I'll probably add it soon (only issue = on grouped bar plots, adding error bars is a bit tricky, so the "coloured data splitting" is probably a bit more annoying to implement, I'll postpone it for the time being).

I'm personally starting to like the syntax: plot(f, df, x) which means do analysis f on the dataframe df regarding some variable x (i.e. hazard, hist, density, expected y given x) and plot the result. It may even be introduced in Plots or StatPlots (together with a set of reasonable analysis). Then one could have an optional population variable meaning: split by this variable, do the analysis, plot mean with errorbars corresponding to s.e.m. If the specific analysis f has a different way of computing confidence intervals, I guess that could be incorporated. This seems like a reasonably general formalism.

Anyway, as soon as I have a bit more time, I'll go on and proceed with the improvements suggested in the first post and make a second PR.

mkborregaard:
I have looked a little more at this, and I think it may have been merged a bit early. I have got some comments I'd like to unfold, but first I'd like to hear if you  @(piever) have any references for the use of this type of plots (split continous data by a categorical variable, do s.e.m. and show that as a continuous ribbon) in modern data analysis? I ask because I have concerns about the robustness of the approach.

piever:
At least in behavioural neuroscience they are pretty common (I can easily find papers using them, but I'm not sure how useful that is because they don't really explain why they use them). The idea is very simple: the categorical variable represent the subjects you are testing and, in order to get a cleaner plot, instead of plotting a family of curves (one per subject) you can simply plot the mean and s.e.m. The curve could be anything: normally it is some column of a database regressed nonparametrically from our variable x (minor digression: to add that I'd like to know what are the good packages in julia for nonparametric regression, so far I've found [this](https://github.com/panlanfeng/KernelEstimator.jl) ), but it could also be some representation of the stochastic variable x (say the cumulative distribution) or pretty much anything.
I'm open to suggestions/contributions and I'm happy to incorporate them in the code, but first I'd need to understand what exactly is the concern: in which sense is this analysis not robust? What alternative(s) do you propose?

mkborregaard:
My concern of the robustness is on whether these shaded errors are often the most appropriate representation of the data. I had a hard time understanding the example with the quakes data and what to get out of it. First it seems that you think `:Stations` refers to a station number that records an earthquake, but it is not - it is the number of stations that recorded the quake. 
```
using StatPlots, RDatasets, DataFrames
quakes = dataset("datasets","quakes")
plot(quakes, :Long, :Lat, :(-:Depth), marker_z = :Stations, ms = :Mag, c = :matter)
plot(quakes, :Mag, :Stations)
```
quakes
figure_1

The point here is not so much that `:Stations` is inappropriate as a grouping variable to do the sem over (I guess you just picked an easily accessible dataset), but that I fear it may be easy to use this function inappropriately.

Another point - the syntax uses `plot(f, df, args..)`, whereas the rest of StatPlots uses `plot(df, f, args...)`, is there a reason for this discrepancy?

piever:
I see. The choice of the dataset is unfortunate (I'm not familiar with RDatasets, I picked a random one and I did think that `:Stations` was some station identifier), but I guess it can easily be solved. First of all we can choose a more appropriate example and add it to the readme, so that people can understand. Also, I'm happy to change the syntax. Just so I understand: would you rather have keyboard argument called say "individuals" (or "subjects" or "population", or something else that you find suitable) that defaults to no splitting (so no shade), and similarly a "split" keyword to split the data that defaults to no splitting, so that the call would be sth like:

`StatPlot.shadederrror(f,:quakes,:Mag; population = :Stations, split = :Deep)`

As for the choice of putting the function before, I though it made more sense for arbitrarily complicated analysis that one may want to run on a dataframe, so that one could use the "do end" syntax.

Anyway I plan on having a second more thorough PR (adding the same population analysis also for scatter and bar plots), so I'd really like to fully understand your feedback and integrate it as much as possible. I guess the correct thing to do is a WIP PR so that everybody has time to give an opinion before merging.

mkborregaard:
Just to be clear, I don't make any decisions around here. I am just a guy like you who thinks StatPlots is cool and would like to see it develop, and I noticed that @tbreloff had asked for some input on the PR. The reason I comment is that I really hope StatPlots will develop into a tool that subtly helps the users make good, appropriate plots of their statistical analysis, rather than just a collection of cool plotting methods.
So the dataset issue can be easily resolved, I only noticed it because I tried to find out what you were doing. The tricky thing with shaded errors is they are only appropriate when the x axis is truly continuous and each x value corresponds to exactly one y value in the underlying data set - which is generally only true for statistical models and relationships where the x axis is time. And for statistical models the error bars will usually be given by the model, or be estimated from data by e.g. bootstrapping. Your density plots provide an interesting exception - it is an appropriate statistical model where it makes good sense to estimate error bars from subpopulations. I just really don't know how common that is, so that is why I asked you for some other examples (also given that the dataset is not appropriate here).
With regards to keyword arguments, that seems to be the case everywhere around Plots that arguments that are not x, y, or z (or the DataFrame) are always keyword arguments, so that would be my personal preference.
With regards to the order of `f` and `df` your argument is sound, but I strongly feel that there should be the same format throughout all StatPlots calls.

piever:
That's a good point. The best I can do is probably add a bunch of symbols that one could use as first argument (instead of explicitly giving the function) that would translate to relevant analysis (also, I am checking whether the x axis is continuous or categorical to make sure the user doesn't do anything silly). I agree that if one started binning data, the error would depend on the binning which is weird, but all the analysis I'm mentioning above (cumulative, density, cumulative hazard, regressing nonlinearly y from x) don't suffer from this drawback.
As for the use of the analysis, it is actually pretty common in fields of research where you have strong variability across subjects (and also you can only have a limited number of subjects) and you want to verify to what extent your plot depends on your sample of subjects.

mkborregaard:
Cool, looking forward to see the examples once you get them into shape! :-)
BTW, with this PR https://github.com/JuliaPlots/StatPlots.jl/pull/29 you can also 
```julia
plot(df,  :(KernelDensity.kde(:Mag))
```
but things such as the `groups` argument aren't working yet, which would otherwise do the same as your populations argument. 

For full disclosure: My secret agenda is that adding smoothers with confidence intervals to plots should not be handled by special userplot functions, but done using normal StatPlots syntax, with type recipes for fitted statistical models including the confidence intervals as just another series type.

piever:
It is actually important to point out that:

`StatPlots.shadederrror(f,:quakes,:Mag; population = :Stations, split = :Deep)`

with an appropriately defined f is in general less nice than just writing:

`StatPlots.density(quakes,:Mag; population = :Stations, split = :Deep)`

and have StatPlots figure out what to do. The only issue in doing that for all statistical recipes is that in general, in the case of a continuous variable, every subject will have a different selection of points on the x axis, making it a bit less obvious how to compute the s.e.m. across subjects. As soon as you allow all of the standard statistical plots to also accept what are the x values used to plot the line, this method can be immediately generalized. Otherwise maybe you could try and figure out what the value of the function would be on that point by interpolating the various line plots, but I don't particularly like the idea.

mkborregaard:
Not sure I understand the point completely - but if you have different values on the X axis, it is more appropriate to use a scatter plot. Because it is important for interpreting the graph to be able to see the distribution on the x axis. 
I agree 100% with the more general point that the call should be for the function, i.e. errors should not require a specific dedicated function to be called.

piever:
It's not a statistical problem but a pratical one: if you split a continuous variable x across say 10 subjects, you get 10 different distributions x1,...,x10. The range is the same, except building a plot for x_1 will choose some points on the x axis (I guess 100 equally spaced from minimum to maximum), so it's not 100% straightforward to compare these different lines and compute the average and an estimate of the variability. As a solution, my PRs only use analysis where you can specify on which points you want your functions to be evaluated (which should have nothing to do with computing the fct, you are just evaluating it) so that you can make sure they overlap. One alternative, which is maybe more error-prone, is to allow the plot to choose points on the x-axis as it wants and when computing the average, interpolate all the line to a common set of points (for example 100 equally spaced from minimum to maximum of the whole distribution). Sorry for explaining that in a bit of a confused way!

mkborregaard:
Ah, I see!

mkborregaard:
The only place where I have experience with that situation is in time-series analysis (i.e. Bayesian skyline plots) - the convention there is to use the union of all points for the x axis; imagine going from the present to now, every time one of the series changes value, the metric (e.g. s.e.m.) is recalculated.

PR #28, opened 2016-12-01
returns a matrix rather than an Any[] from handle_dfs

mkborregaard:
Solves #24

### DIFF
--- a/src/dataframes.jl
+++ b/src/dataframes.jl
@@ -8,7 +8,8 @@ end
 # if it's an array of symbols, set the labels and return a Vector{Any} of columns
 function handle_dfs(df::AbstractDataFrame, d::KW, letter, syms::AbstractArray{Symbol})
     get!(d, :label, reshape(syms, 1, length(syms)))
-    Any[collect(df[s]) for s in syms]
+    ret = Any[collect(df[s]) for s in syms]
+    hcat(ret...)
 end
 
 # for anything else, no-op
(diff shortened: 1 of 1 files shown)


src/dataframes.jl
```diff
@@ -8,7 +8,8 @@ end
 # if it's an array of symbols, set the labels and return a Vector{Any} of columns
 function handle_dfs(df::AbstractDataFrame, d::KW, letter, syms::AbstractArray{Symbol})
     get!(d, :label, reshape(syms, 1, length(syms)))
-    Any[collect(df[s]) for s in syms]
+    ret = Any[collect(df[s]) for s in syms]
+    hcat(ret...)
```
tbreloff: This feels like the wrong solution... build a matrix just to break it apart again?  Plus what if the columns are different types?  Sorry I can't focus on this right now

mkborregaard: It is that the code creates a row Matrix{Vector{Float}} of length `n`, and Plots don't like that - this code makes it a Matrix{Float} with `n` columns, which is what Plots likes.
I'll remind you later if you don't have time to look at it now :-)

tbreloff: Oh ok I misread... then the better approach would be: `Any[collect(df[s]) for i=1:1, s in syms]`  Though I thought Plots should be able to handle a `Vector{Any}` properly.

mkborregaard: Yes, Plots, handles a Vector{Any} well, but that's the issue: this code `Any[collect(df[s]) for s in syms]` creates a Vector in julia 0.4 but a 1-row Matrix in julia 0.5, which is what causes the error. Replacing hcat with `vec` works.

tbreloff:
What sorts of commands have you tested this with?

mkborregaard:
Just 
```julia
using DataFrames; df = DataFrame(a = randn(100), b = 1:100, c = randn(100))
plot(df, :a, :c)
plot(df, :a, [:b :c])
```
I think you could say that is not comprehensive testing. I am happy to do more tests.

mkborregaard:
Did you see my explanation for this? - I added them to the outdated comments.

PR #29, opened 2016-12-06
Expressions with symbols for x and y

mkborregaard:
I added the ability to use expressions for the `x` and `y` variables as well, rather than just things like `markersize`. With this PR you can do things like 
```julia
using DataFrames; df = DataFrame(a = 1:100, b = randn(100), c = randn(100))
scatter(df, :a, [:c :(log(abs(:b)))], ms = :(2*abs(:b)), color = [:red :blue])
```
and have it work. I have also cleaned up the code (it is still somewhat messier than before the addition).

There are still some issues with this:
1. gr() fails when passing x or y as expressions - pyplot() works though.
2. because i pass a Vector{Any} rather than a Matrix for the x and y variables, and use the same functionality for things like `markersize`, I cannot have `ms = [:a :b]` work (this seems to be an issue already in my previously merged PR, sorry). So it is marked WIP, though it generally works.


tbreloff:
@mkborregaard What is your plan with this PR?

mkborregaard:
I'd like to solve the two issues I mention above and merge it. I often miss this functionality. It's been a while since I looked at it, but let me try to see if I can quash it over the weekend.

mkborregaard:
This should work now and is ready for review.

Issue #3, opened 2016-07-11
cornerplot


tbreloff:
See: https://github.com/JuliaPlots/PlotRecipes.jl/issues/1#issuecomment-226729872

cc: @mauro3 (do you want to submit a PR?)

mauro3:
Will do.

mauro3:
Here the readme picture:
![cornerplot](https://cloud.githubusercontent.com/assets/4098145/16762241/f924c8ac-4823-11e6-8720-83025606b5e1.png)

tbreloff:
i've been fixing this up.  new screenshots:

![tmp](https://cloud.githubusercontent.com/assets/933338/17787905/2e161b7c-6559-11e6-937a-20209301905b.png)

![tmp](https://cloud.githubusercontent.com/assets/933338/17787917/3af9b0e2-6559-11e6-8613-c177efb0b038.png)

mauro3:
Cool! 

One thing which would be nice to have at some point is the have combined contour-hist-scatter plots, like so: http://corner.readthedocs.io/en/latest/pages/quickstart.html

tbreloff:
Can't you do that already with corrplot? (Well, not with "compact")

mauro3:
I mean this plot:
![p](https://cloud.githubusercontent.com/assets/4098145/17809692/ae02a62e-6618-11e6-9c36-7cc406893a02.png)
How would you do that?

tbreloff:
I guess I don't understand what it is. Is it a partial hist2d, but the
individual points are drawn when there's only a few?  And a contour
overlayed?  Certainly possible but might be slightly annoying to make.
You'd want to draw the hist2d as shapes I think. So the recipe would make 3
series... a shape, a scatter, and a contour.

mauro3:
(I don't care about the histogram.)  But the contours are good as the scatter-plot is too dense in the middle to discern anything.  Anyways, just an idea for sometime-in-the-future work...

tbreloff:
Oh. Then that would be pretty straightforward

mauro3:
I'll ping you when I need/want it!

mauro3:
I think it would be good to have x/y-tick labels along the top and left too.  With many variables it gets tricky to see where histograms are.

mauro3:
Also using the `title` keyword sets a title on all subplots.

PR #30, opened 2016-12-10
[WIP] Add population analysis with error bars across population for continuous plots

piever:
After the discussion in #27 I've tried to come up with a consistent way to insert error bars for continuous plots representing standard error across a population. I've chosen the dataset

`school = dataset("mlmRev", "Hsb82")`

where a bunch of school are sampled and in each school one can get a distribution of values. Then it is natural to run a given analysis per school (e.g. cumulative of a value, a density plot, regress y from x possibly nonparametrically) and then plot the mean of this analysis across schools with a shaded standard error across schools. I've added the possibility of doing that (for shaded error plots, bar plots and scatter plots) giving the analysis function, the dataset and the arguments necessary for the analysis function. I've added the possibility to split data with the keyword group. The keyword "across" means which variable represents the population to compute s.e. The others keywords are transmitted to the plot. The legend is computed automatically, but can be modified (a vectorial keyword will be cycled across the different split data). I've also added a few functions that could be useful to run this analysis (one that computes density, one for the cumulative and one for locally-linear regression), both for a continuous and categorical x axis.

Example:

screenshot 2016-12-10 19 23 21

The same type of analysis can be done with bar plots (or also scatter plots):

screenshot 2016-12-10 18 55 32

I've added a different package (KernelEstimator) because it allows you to compute these statistical analysis specifying what points to have on the x-axis (when I split the data across the population I need the same x-values to compare y-values and get a standard error) and it has an implementation of nonparametric regression, which I couldn't find on KernelDensity. As I mentioned in #27 , the alternative is to not require that and, should different subjects have different x-values, one could just choose common x-values at the end and interpolate what the values of the various subjects would be at the common x-values. I'm working on that right now.

I'd be very happy to get suggestions on what is the best syntax and what is the best way to incorporate this stuff in the plots.jl ecosystem: my intuition is that, if one adds the interpolation option, it should be reasonably easy to get this kind of error-plots using Plots.jl features in a possibly smarter way (e.g. a keyword that does the same as "group" to split the data but instead of plotting separately finds mean and s.e.m. and plots correspondingly). However I don't think I understand the Plots.jl machinery well-enough to implement that kind of solution, so I'd be happy to get directions/help/code.


mkborregaard:
I am happy to have a look at this as soon as I have a moment, but consider this: would it not be nice to think of the errors as simply another series to add to a plot, or something that could be lifted from a Fit object? I have wanted to do a PR on that for a long time, so I really hope we can consider this possibility before moving StatPlots off in the direction of this PR.

I'll open an issue this week detailing my ideas for support of statistical objects and error bars in StatPlots, so that everyone can review it.

piever:
That's why I put the "WIP" label: so we can consider all the different options. I though the best way to start a discussion was to propose one way and then see what other people think. I do believe that if one knows Plots.jl better than I do, maybe there's an even easier way of doing this and I'm looking forward to seeing it :)

mkborregaard:
Hi @piever I haven't forgotten I promised you feedback on this :-) It is a nice job on the code, but I have been wrecking my brain on how these recipes could fall naturally into the plots ecosystem. Here are some things to consider, I am happy to discuss any of this:

1. I agree with what you said in #27 that `density(quakes,:Mag; population = :Stations, split = :Deep)` is better than `shadederror(f,:quakes,:Mag; population = :Stations, split = :Deep)`. The density plot is what the user wants, whereas the mean-and-error is just a way to show the variation. I think it would be better to make `density` and `cumulative` the user plots.

2. I am not so keen on importing bar and scatter from Plots and giving them new keywords (it becomes extra confusing as `group` is already a keyword in Plots).

3. My main worry is that it becomes a little too magical for someone not very clear on what is going on. Might you consider separating the analysis from the plot? This would make it a two-step process: using a function to create a grouped_summary object, and plotting that object. For the first step you could have a function that splits the dataset, applies some function and summarizes the variation using eg s.e.m. or confidence intervals, and gives the result as a custom type. For the second step you could have a type recipe for that type of object - then the series type (line, bar or scatter) and error type (error bars, shaded polygons) could be specified in the plotting or automatically based on attributes of the object. I think that would make the code more modular and extensible, and more transparent and intuitive for the user.

piever:
Actually, 3 seems like the best solution (also, ideally this wouldn't be the only case case where a model has some uncertainty and one wants to visualise it with a shade: simple linear regression with one predictor is another obvious example). I'll be playing a bit with recipes to see if I can get it to work, but actually it should lead to simpler code than what I wrote so far (as I had to basically reimplement group and things like that).

mkborregaard:
> simple linear regression with one predictor is another obvious example
Exactly! I have in fact recently opened a PR on GLM to allow `predict`  to generate confidence intervals (https://github.com/JuliaStats/GLM.jl/pull/171) that can be plotted as shaded errors. I see some nice further possibilities in this - eg regressions with a categorical and a continous variable can be plotted as several colored lines with shaded errors, a catogorical anova as a bar plot with contrasts and significance levels etc.
The idea is to keep the statistical analysis and the plotting separate - to put the analysis completely in the hands of the user, but allow extensive support for intuitive plotting of an array of statistical objects.

piever:
The only trick seems to be that one has to pass the analysis function as a keyword argument. As far as I understand, what you tried on gitter can't really work:

`plot(quakes, kde(:Mag), group = :Deep) `

because plots.jl will try to split the second column according to quakes[:Deep], but now the second column doesn't even have the same length, it is whatever kde outputs, whereas instead you should be splitting the dataset. So a specific type series (called for example population analysis) with a signature like

`populationanalysis(quakes, :Mag, group = :Deep, analysisfunction = density)`

and now I believe that the first thing happening will be the splitting of the dataset and then you can run whatever analysis you want, based on keyword analysisfunction.

With this (plus generalizing group to work with more than one label) I think one could get the same functionality as I have now but in a smarter way, I'll try it as soon as I have enough time and will keep you posted.

mkborregaard:
No, that's right, the `group` approach I tried on Gitter cannot work, because the kde function is not elementwise on the elements of x. That is how the function interface to Plots work by convention (the implementation is `map(f,x)` in fact). `kde` is whole-vector, along with a few other methods such as generating histograms, and because of that I don't think it should use the function-passing syntax of Plots. 

Instead, I think most users would find it intuitive to have a `densityplot` and a `cumdensityplot` function, that follows the same syntax formula as `histogram` (which should be easy to do with parts of your code). IMHO, though, they should not take 3-4 arguments that are specialized for one particular way of generating error bars. That should be taken care of in an additional step.

The issue with the error function is that both the mean line and the errors are created from the subsetting, i.e. the mean is not in the dataset. That is why I think it could be nice to have two steps (non-working pseudocode):
```julia
by_Sx = groupapply(cumulative, school, on = :Mach, by = :Sx, summarize = :sem) #no problem with having the function first here
plot(by_Sx, error = :ribbon)
```
So in principle what we have is a function for a statistical analysis that can be used in any further analysis, and a plot function that follows the generic syntax. I think this, though requiring two lines, will give a more intuitive workflow for most people with less requirement for looking into documentation.

piever:
Actually I guess I could easily get that pseudocode to work with code that I have already. I still have one doubt however as to what should be the type of outcome from groupapply.

Option 1)
A dataframe, with one column for x values, one for y values, one for error, one group membership. In this case the plot call would be:

```
by_Sx = groupapply(cumulative, school, on = :Mach, by = :Sx, summarize = sem)
shadederror(by_Sx, :x, :y; shade = :error, group = :group)
```

Disadvantage: wouldn't work with groupedbar! But maybe one could also fix groupedbar to work with group.

Option 2) 
A specific type where by_Sx.x is the vector of x points, by_Sx.y is the 2D array of y points (one axis is x coordinate, the other the group), by_Sx.error is the 2D array of errors, 
Disadvantage: x axis for different groups may not have same number of points, so it would be a bit artificial to force them to. Also maybe the dataframe format is more handy if one wants and continue the analysis.

What do you think?

piever:
Extra issue with giving a dataframe: shadederror doesn't play with grouping as well as say ribbon or err keywords, so that this:

```
by_Sx = groupapply(cumulative, school, on = :Mach, by = :Sx, summarize = sem)
shadederror(by_Sx, :x, :y; shade = :error, group = :group)
```

wouldn't work whereas this would (but not in plotlyjs):

```
by_Sx = groupapply(cumulative, school, on = :Mach, by = :Sx, summarize = sem)
plot(by_Sx, :x, :y; ribbon = :error, group = :group, fillalpha = 0.5)
```

If you know how to make shadederror behave just like ribbon, that'd be nice :)

Also, I have a technical question. If we go for the specialized type, say by_Sx::GroupedError then one would want `plot(by_Sx, style = :line)` to return `shadederror(by_Sx.x, by_Sx.y; shade = :by_Sx.error, label = by_Sx.label)` whereas `plot(by_Sx, style = :bar)` would return `groupedbar(by_Sx.x, by_Sx.y; err = :by_Sx.error, label = by_Sx.label)` and so on. However while it's straightforward to do it importing the function plot and expanding it with the new type by_Sx, I'm not quite sure what's the elegant way to do this with recipes in a way that plays well with the grouping of the data (because, as far as I understand, the type recipes happen before the data splitting). Could you give some example code that would do that?

mkborregaard:
Thanks for all the considerations - I'll answer on Tuesday

tbreloff:
If you create a user recipe for a new type GroupedError, then you can do
all that logic in the recipe for GroupedError. ie:

bar(grperr) // return GroupedBar object from recipe

User recipes are processed recursively, so you can return another plottable
object, with whatever logic you want. Just check the seriestype inside the
recipe and change it however you want.

And in general I much prefer option #3 for something like this... Create
some custom object for complex statistics on a dataset, then make a user
recipe for that object. The functionality isn't generic enough to be adding
all these keywords. (IMO)

mkborregaard:
Glad to hear you agree @tbreloff :-)
In my opinion, a statistical plotting image is not one that does the statistical analysis and plot it - it is one that helps the user visualize the statistics that she is doing! And specialized type recipes for statistical objects is the way of doing that.

@piever about your first question, I meant number 2 - create a specialized object, then a type recipe for that (i.e. call it using `plot` - in general, I think `shadederror` isn't the best name for this functionality).

The reason I hesitated instead of responding immediately - was the thought "could this be done with a type recipe for GroupedDataFrame that split each group into series?". You could `by_Sx = groupby(schools, :Sx)`, then `plot(by_Sx, KernelDensity.kde)`, which should create a density plot with two series. There could then be an extra argument called `summarize` that takes either `:none` or a tuple of functions for the central trend and variation, e.g. `(mean, sem)`. Just a thought, I think that would be both modular and intuitive (but I haven't thought it all the way through).

piever:
Thanks for the feedback. I changed it to work in the most general scenario (with mean and sem as standard analysis, but it can be changed). I've added a recipe that takes this group error type and depending on the seriestype does line plot with shadederror or scatter with errorbar or groupedbar with errorbar. The only thing that remain to be decided is what type this groupederror should be. I'm of the opinion that the most sensible would be for it to be a dataframe (either grouped or with a column indicating group membership) and I could add a small function that reshapes a dataframe so that it can work with groupedbar.

mkborregaard:
Can you update the PR to show your changes?

piever:
Still need to fix this decision of a type that works easily both for line plots and groupedbar, but I should be able to push by today.

mkborregaard:
Just make your own.

piever:
This one should be working, and I think I've incorporated more or less all the feedback. There is a groupapply function that splits the data across kw "group", then applies "summarize" (default is (mean,sem), but you can put any pair of functions. Also you have the option shared_xaxis whether you want a common x axis for all the split data (which is required for groupedbar but not recommended for shaded error). Also, for the local regression I'm using Loess (like Gadfly does) instead of KernelEstimator because it seems better maintained.

Example use:

```
using DataFrames
using RDatasets
using Plots
using StatPlots
gr()
school = dataset("mlmRev","Hsb82");
grp_error = StatPlots.groupapply(:cumulative, school, :MAch; across = :School, group = :Sx)
plot(grp_error, line = :line)
```
screenshot 2016-12-19 12 28 27

Keywords for loess or kerneldensity can be given to groupapply:

```
df = StatPlots.groupapply(:density, school, :CSES; bandwidth = 1., across = :School, group = :Minrty)
plot(df, line = :line)
```

screenshot 2016-12-19 12 34 43



The bar plot

```
pool!(school, :Sx)
grp_error = StatPlots.groupapply(school, :Sx, :MAch; across = :School, group = :Minrty, shared_xaxis = true)
plot(grp_error, line = :bar)
```
screenshot 2016-12-19 12 29 52

Only problem left (but independent of this PR): with plotlyjs() groupedbar doesn't work well with errorbar, I'll open an issue for that.

piever:
I've been playing a bit with the code today on my data. It actually seems to work reliably. There's only one thing last design decision on which I'm not 100% sure, which is how to choose between smooth and discrete analysis (i.e. computing density by checking frequency of each individual value of x variable vs doing a kernel density analysis). In the currently uploaded code, if the column is a pooleddataarray it does the discrete analysis, otherwise it does the smooth one. If one tries to give a non-pooled array of a discrete type(say Int64 or Bool or String), the code as of the last pushed version will actually give an error. I'm not sure which is the recommended behaviour, as I can imagine some users having an array of integers that maybe has such a large range that they want to treat as a continuous variables, so I guess there are 3 options:
- Correct so that it will work also for integers as continuous variable
- No longer check on pooled vs non-pooled but add a keyword to decide on the type of analysis
- Leave it as is

which one do you think is best? I think I'm in favor of number one, but maybe also number 2 as its advantages, in case one is undecided on which analysis is best. Maybe I could add a keyword to override the decision (if you don't assign the keyboard it chooses according to pooled not pooled, otherwise it follows the keyboard).

The other things that I'm not worrying about is missing data, but I guess I'll wait that the situation with missing data and DataFrames clarifies before working on that.

src/bar.jl
```diff
@@ -43,19 +43,18 @@ grouped_xy(y::AbstractMatrix) = 1:size(y,1), y
...
+    fr = zeros(nr, nc)
+    for c=2:nc
+        for r=1:nr
+            fr[r,c] = y[r,c-1]
+            y[r,c] += fr[r,c]
         end
+    end
         fr
     else
-        get(d, :fillrange, 0)
```
mkborregaard: remind me why you are throwing away the fillrange value here for non-stacked bars?

src/groupederror.jl
```diff
@@ -0,0 +1,137 @@
+type GroupedError
```
mkborregaard: please type the fields with `::`

src/groupederror.jl
```diff
@@ -0,0 +1,137 @@
...
+        valid = !isnan(mean_across_pop) & !isnan(sem_across_pop)
+    end
+    if xaxis == false
+        return xvalues[valid], mean_across_pop[valid], sem_across_pop[valid]
+    else
+        return xvalues, mean_across_pop, sem_across_pop
+    end
+end
+
+function groupapply(f::Function, df, args...;
```
mkborregaard: docstring please

src/StatPlots.jl
```diff
@@ -24,5 +25,9 @@ include("hist.jl")
 include("marginalhist.jl")
 include("bar.jl")
 include("shadederror.jl")
+include("analysisfunctions.jl")
+include("groupederror.jl")
+
+
```
mkborregaard: you should export the groupapply function

src/groupederror.jl
```diff
@@ -0,0 +1,137 @@
...
+
+get_axis(column::PooledDataArray) = sort!(unique(column))
+get_axis(column::AbstractArray) = linspace(minimum(column),maximum(column),200)
+
+# f is the function used to analyze dataset: define it as nan when it is not defined,
+# the input is: dataframe used, points chosen on the x axis, x (and maybe y) column labels
+# the output is the y value for the given xvalues
+
+function get_mean_sem(trend,variation, f, df, population, args...; xaxis = false, kwargs...)
+    # define points on x axis
```
mkborregaard: better `xaxis = :auto`

src/groupederror.jl
```diff
@@ -0,0 +1,137 @@
...
+    if xaxis == false
+        xvalues = get_axis(df[args[1]])
+    else
+        xvalues = xaxis
+    end
+
+    if population == []
+        mean_across_pop = f(df,xvalues, args...; kwargs...)
+        sem_across_pop = zeros(length(xvalues));
+        valid = ~isnan(mean_across_pop)
```
mkborregaard: why `~` not `!`?

src/groupederror.jl
```diff
@@ -0,0 +1,137 @@
...
+    else
+        # get mean value and sem of function of interest
+        splitdata = groupby(df, population)
+        v = Array(Float64, length(xvalues), length(splitdata));
```
mkborregaard: It would be cool to define this mean_and_sem behaviour in a separate function tha tworks on a `GroupedDataFrame` (the type of splitdata), has the same argument structure as the full function (minus `population`), and is called here. That could be exported as well, then a user could call this directly on a `DataFrame` that he had already split himself with `groupby`.

src/groupederror.jl
```diff
@@ -0,0 +1,137 @@
...
+        for i in 1:length(splitdata)
+            v[:,i] = f(splitdata[i],xvalues, args...; kwargs...)
+        end
+        mean_across_pop = Array(Float64, length(xvalues));
+        sem_across_pop = Array(Float64, length(xvalues));
+        for j in 1:length(xvalues)
+            notnan = !isnan(v[j,:])
+            mean_across_pop[j] = trend(v[j,notnan])
+            sem_across_pop[j] = variation(v[j,notnan])
+        end
```
mkborregaard: This way of dealing with `NaN` means that confidence bands will all taper off towards the ends, right? I think that is valid enough, but might be more conservative to skip those x values with NaNs completely

src/groupederror.jl
```diff
@@ -0,0 +1,137 @@
...
+        return groupapply(analysisfunction, df, args...; kwargs...)
+    else
+        return groupapply(_locreg, df, args[1], s; kwargs...)
+    end
+end
+
+groupapply(df::AbstractDataFrame, x, y; kwargs...) = groupapply(_locreg, df, x, y; kwargs...)
+
+@recipe function f(g::GroupedError)
+    if !(:seriestype in keys(d)) || (d[:seriestype] == :line)
```
mkborregaard: should probably be seriestype `:path`, not `:line`

mkborregaard:
To your question @piever I suggest promoting integers to Floats, keeping the test for PooledArray, and adding a keyword.

src/bar.jl
piever: It's a fix to: #32 In case of errorbars, the fillrange to 0 also gets applied to the error series and creates those messy things of the issue (on PlotlyJS). Without any fillrange it seems to work just fine.

src/groupederror.jl
piever: Ups, apparently programming in matlab left a mark on me. I guess the two are equivalent, but I'll be consistent with the rest of the world and use `!`

src/groupederror.jl
piever: Actually with real data these plots tend to have huge error bars towards the end, which is somewhat annoying, but it could also be that one has a lousy subject with a weird range and you don't want that to prevent you from looking at most of your data. As of now I require at least 2 subject to be not NaN to plot the mean and sem but that could be another keyword (in terms of number of subjects or fraction of subjects required to consider it a "plottable" point). In principle one could even think to show with some vertical lines or sth else where is the majority of the data (say from 0.05 to 0.95 quantile).

piever:
Thanks a lot for the thorough feedback!
As for your doubt, the barplot actually only makes sense when the x variable is discrete. In the discrete case, _locreg is conditional expectation of y given x (so for every value of x, I take the data points with that value and get the average of y). Maybe _locreg is a bit of an unhappy name and one should call it _xyplot or something, so that the name also makes sense in the discrete case.
So in the example, the way the first blue column is computed is as follows: take only the data points corresponding to female subjects with `:Minrty = No`. Then for every school see what's the average `:MAch`, then take the mean and sem of this quantity across schools. Basically here the x variable is just another way of splitting the data. Actually, if instead of doing `across = :school`, one did `across = :row_number` (or any other variable that is different for different datapoints) we would recover the classical bar plots (though it may be an inefficient implementation).

src/bar.jl
mkborregaard: Hmm I worry what that may have of unintended consequences. Does `get(d, :fillrange, nothing)` work?

src/groupederror.jl
mkborregaard: No I think it is fine to leave it as is.

src/bar.jl
piever: Yes, that also works. I'll go for `get(d, :fillrange, nothing)` to stay as close as possible to the original design then.

tbreloff:
@mkborregaard I'm leaving this PR in your hands... merge it whenever you're happy.

piever:
@mkborregaard : Added the changes and cleaned up the axis selection a bit. Now I use the keyword axis_type and if it's `axis_type = :auto` I check if it's a categorical column. I've also removed the need to specify the shared axis to do a bar plot, it does it automatically in the recipe. Let me know if you have any more feedback!

Best,
Pietro

src/analysisfunctions.jl
```diff
@@ -0,0 +1,37 @@
+function _locreg(df, xaxis::LinSpace, x, y; kwargs...)
+    model = Loess.loess(convert(Array{Float64,1},df[x]),convert(Array{Float64,1},df[y]); kwargs...)
```
piever: Actually, I can imagine that some of the columns could be integers with a very wide range (e.g. some time given in milliseconds) and the user would still want to use Loess regression which however doesn't accept arrays of integers, so I thought it was better to leave it as is.

README.md
```diff
@@ -121,3 +121,40 @@ groupedbar(rand(10,3), bar_position = :dodge, bar_width=0.7)
...
+
+## groupapply for population analysis
+There is a groupapply function that splits the data across a keyword argument "group", then applies "summarize" (default is (mean,sem) but the user can put any pair of functions) to a given analysis (density, cumulative and local regression are supported so far, but one can also add their own function). The local regression uses [Loess.jl](https://github.com/JuliaStats/Loess.jl) and the density plot uses [KernelDensity.jl](https://github.com/JuliaStats/KernelDensity.jl). In case of categorical x variable, these function are computed by splitting the data across the x variable and then computing the density/average per bin. The choice of continuous or discrete axis can be forced via `axis_type = :continuous` or `axis_type = :discrete`
+
+Example use:
+
+```julia
+using DataFrames
+import RDatasets
+using Plots
```
mkborregaard: You don't need `using Plots`

src/groupederror.jl
```diff
@@ -0,0 +1,178 @@
...
+        return get_mean_sem(trend,variation, f, splitdata, xvalues, args...; kwargs...)
+    end
+end
+
+"""
+    groupapply(f::Function, df, args...; axis_type = :auto, across = [], group = [], summarize = (mean, sem), kwargs...)
+
+Split `df` by `group`. Then apply `get_mean_sem` to get a population summary of the grouped data.
+Output is a `GroupedError` with population summaries, then can be plot using `plot(g::GroupedError)`
+Seriestype can be specified to be `:path`, `:scatter` or `:bar`
```
mkborregaard: There are a few typos here. Also, the `summarize = (mean, sem)` argument seems to be at odds with "Then apply `get_mean_sem`", as that would change with different `summarize`, right?

mkborregaard:
Overall, two questions:
1. It is not clear what the `across` argument does - is it needed, and always preferred? Should the default not be to do it across all observations?
2. About the _locreg for bar plots - why not just compute the mean and sem for consistency? Am I missing something here?
The more I think about it, the barplot functionality seems to be a separate thing from the path functionality. It would be nice if groupedbar was a series recipe, then you could define this as an alternative (and much more convenient) way to specify a groupedbar plot.

piever:
1) `across` tells it wich variable to use to split the data to compute mean and sem. As of now, if `across` is not given, it will simply not compute any error.
You have a good point that in the case "plot y as a function of x" (and I believe only in this case) where x is a discrete variable there is the obvious interpretation of using the sem of y given x as error (now to get it you'd need to do `across = :row_number`, or some other variable that is always different for different datapoints, which is a bit silly). I can certainly include that. Also, in the case of regressions (e.g. Loess, or your linear fit, which can also eventually be added) one could maybe get some notion of error and plot it (but I don't know what's an estimate of error for the Loess or how to get one, maybe that can be for a later edit). I guess you could in the future add a "linreg" function here, where the error is the sem across individuals if you give `across` and an estimate of error of the regression line otherwise.

2) Question 2) is related: once I make the correction  for the `across` default, then the bar plot would compute mean and sem if no `across` is given, else if you give some `across` (i.e. `:School` ), it would compute mean and sem across that variable (i.e. schools). Given that only the plotting style is different, but the analysis is the same, I guess that at least for the time being this can stay here. If in the future you find a way to put it as an other method of groupedbar, we can decide which option is best, or maybe even leave both, but at least for now users have a more convenient way of doing bar plots.

I'll implement your suggestions and the `across` improvement before merging then.

mkborregaard:
That sounds great - can you make sure `:across` is described carefully in the readme? It'd be great with a comprehensive readme (it is already nice what you have) for this operation, as it is fairly complicated for an outsider :-) Also, unless you think other functionality should use the analysisfunctions.jl I think it might as well just be in the same file as the plotrecipe

mkborregaard:
Oh and thanks for holding out on this one - you've done an amasing job editing and working with this, I think the functionality is really cool now. :-)

piever:
In the end I think the most general way is two give 3 options: 
1) `across` is a variable name (or an array of variable names) and the summary is done on that
2) `across = true` (maybe default?) and the error is computed in some way specified in the analysis function (i.e. sem of all the ys corresponding to a given x value, or error of linear regression when we add linreg). If there is no implemented way of getting the sem from the function, it can be computed with nonparametric bootstrapping (i.e. for the Loess regression).
3) `across = false` (maybe default? To be used if bootstrapping is too computationally expensive). No error is computed or displayed.

In any case it needs to be well documented because it is a non-trivial keyword.

Actually I think your feedback was pretty useful to improve this PR! I've sadly realised that the "review" interaction is much more productive when publishing code than when trying to publish my research :-)

mkborregaard:
Great to hear! I think the across parameter should not be so complicated. across can take a variable name and act as it does now. If not specified it works over all observations.

piever:
That'd be ideal, but the main issue is that for some of the functions it doesn't make sense. For example if the x axis is continuous you can't draw the curve with one observation, so you can't really split by observation (hence the suggestion to use bootstrap). For density and cumulative, in the discrete case, I guess you can do it individually per observation (getting density and cumulative of delta functions) and the average would be correct, and the sem would also be informative. Still in general I think that there could be some analysis where splitting by observation and taking the mean has little to do with doing the analysis on all the observations together (or, is the same but is way less efficient) so I would be in favor of giving the option of not splitting and simply not getting the error. I'm trying to come up with a way of letting the user do that that is not too error-prone.

piever:
By thinking a bit more, given that we plan on adding built-in errors (see linreg, but also for locreg) it would maybe more sense to have a more general keyword:
`compute_error `

It could have as options: `compute_error = (:across, column_name)` to work with the across keyword as this PR is currently doing, `compute_error =:across` would be across all observations, `compute_error = :none` to have a faster computation, or `compute_error = :built_in` if the user prefers to compute its own version - or for linreg we could give one ourselves- or `compute_error = :bootstrap` for example to run the bootstrap analysis (we don't need to implement all of them right now). In this case I would maybe propose `compute_error =:none` as default, but I guess that's up for debate.

What do you think?

REQUIRE
```diff
@@ -6,3 +6,4 @@ StatsBase
 Distributions
 DataFrames
 KernelDensity
+Loess
```
tbreloff: Sorry I missed this comment.  Yes adding dependencies should not be done lightly, but this is probably acceptable.

src/StatPlots.jl
```diff
@@ -9,11 +9,15 @@ using Distributions
...
 
 import KernelDensity
+import Loess
 @recipe f(k::KernelDensity.UnivariateKDE) = k.x, k.density
 @recipe f(k::KernelDensity.BivariateKDE) = k.x, k.y, k.density
 
 @shorthands cdensity
 
+export groupapply
+export get_summary
```
tbreloff: do we need to export `get_summary`?

piever:
@mkborregaard  I've managed to add everything I wanted. Now `compute_error = :none` doesn't compute any error. `compute_error = :across` is across all observations. `compute_error = (:across, col_name)` is across the variable `col_name`. `compute_error = (:bootstrap, n_samples)`, generates `n_samples` fake datasets using nonparametric bootstrapping and then summarizes them. Everything is explained in the README. Addressed your concerned about the naming of `get_mean_sem`. As far as I'm concerned, I think it's ready to go.

@tbreloff I don't have strong opinions on that. The actual plotting is done as in the README:
```julia
grp_error = groupapply(:cumulative, school, :MAch; compute_error = (:across,:School), group = :Sx)
plot(grp_error, line = :path)
```
and in practice I never really use it. Is it bad to export it?

src/StatPlots.jl
piever: I don't have strong opinions on that. The actual plotting is done as in the README:
```julia
grp_error = groupapply(:cumulative, school, :MAch; compute_error = (:across,:School), group = :Sx)
plot(grp_error, line = :path)
```
and in practice I never really use it. Is it bad to export it?

tbreloff: In this case it doesn't seem necessary, but I admit I don't actually know when you'd call it.

piever: It was proposed by @mkborregaard. The idea is that in the normal scenario, you would have a dataframe that you'd split according to a column and the summarize the split data  (for example getting mean and sem). The use case would be somebody having obtained a grouped dataframe already by some other means and wanting to get the summary, which is not possible with only `groupapply`.
The pipeline is: `groupapply` calls `get_summary` who calls the analysis function on the subdataframes. `groupapply`, `get_summary` and the built-in analysis function have docstrings. The docstrings of `groupapply` refer to `get_summary`, so maybe the user is expecting that it should be exported.

piever:
Slight off-topic, but I'm not sure where to ask: I'm making a jupyter notebook with many examples from this library (mainly for my lab-mates). In particular I'll show some examples similar to the README but in more detail and I'll also show how to use the functionality of `groupederror` together with the Interact.jl package. Basically you can do a simple `@manipulate` adding a couple of sliders for smoothing paramaters, and 3 or 4 more symbol variables representing the choice of the variable on the x axis, the choice of the analysis (i.e. cumulative, density or the name of a variable to plot on the y axis as fct of x), some grouping variable and how to compute the error. 
With a few lines of code you can actually produce some sliders and toggle buttons that will allow you to do interactive exploratory data analysis (i.e. look at more or less all the variables and their relationships) on your dataset with a few clicks (I think R has a GUI that does that, or so I'm told).
I was wondering where I should upload this notebook? Should I make a separate repo and add the link or there is an appropriate place? I was thinking maybe ExamplePlots.jl, even if it's not a Plots example but a StatPlots example.

mkborregaard:
Thanks for all your work on this - I am at a conference in the US currently but will look it over as soon as I get a chance.

tbreloff:
ExamplePlots would probably be the right home for something like this.

piever:
I see, then I guess I'll try and do a PR there shortly (to be merged after this one).

src/StatPlots.jl
piever: Actually, on a second thought I'm in favour of keeping it also due to performance reasons: for  large datasets it may be inconvenient to have to group it every time. In the implementation of `compute_error = :bootstrap` I see a seizable performance improvement by working with GroupedDataFrames directly rather than putting everything in a big DataFrame and then applying groupby.

piever:
@mkborregaard I've been testing the framework more extensively and 2 minor issues came out:
1) So far I compute the axis separately for the different groups. Sometimes, if the ranges are very different, it looks a bit silly to have for example two cdfs ending in different places. What I can do is to keep this behaviour as default and add the option of having a common x axis / giving the x axis by hand.
2) Sometimes (in the dataset that I'm testing, but it never happened in my own data), after splitting the dataset there's not enough data to run some slightly more sophisticated analysis (i.e. Loess or KernelDensity) so it gives an error. Should I add some try/catch to just give NaNs for the subjects not having enough data to run the analysis or should I just let it error and let the user figure it out? My feeling is that if each subject doesn't have enough data, this approach is ill-advised anyway and one should try hierarchical models.

src/StatPlots.jl
mkborregaard: Yes exatly, that was the idea. I think it should be renamed to something more specific - "get_summary" is too generic a name to put in the user's workspace for a very specific function used for one type of plot only

mkborregaard:
1. Agree, leave as is.
2. Let it error and let the user figure it out. Or alternatively catch the exception, write a suitable error describing what went wrong, and re-throw the exception.

mkborregaard:
Other than the name change of `get_summary` (and possibly the suggested error handling), I am ready to merge if you still think it is ready  @piever . Thanks for the great and focused work on this!

piever:
I have a very minor restructuring of the code (to allow the user to give a axis to get_summary) and I think it's good to go. What name do you propose for get_summary ? Maybe get_groupederror?

mkborregaard:
Yes that could work

piever:
Changed the name of get_summary, I think it's ready now.

mkborregaard:
Nice, great work!

PR #34, opened 2017-01-19
NullableArray DataFrames

mkborregaard:
Adds support for the new DataFrames type.
Will replace `Number`s that are `Null` with `NaN`, and fail if any non-number arguments contain Nulls (so far).


mkborregaard:
@tbreloff how about this for a first pass?

mkborregaard:
When would a non-numeric Array containing Nulls be passed to `plot`? The only cases i can think of is for `group` (which is handled explicitly), and `hover` (where it would be most appropriate to replace the String or Symbol with "").

Here is more generic but quite clunky approach to fix non-numeric arguments with nulls - let me know what you think of it: 
1. parse `values(d)` and `args` (including any `Expr` types) for `Symbol`s and put those in a `set`.
2. put the `DataFrame` names in another `set`
3. make a new df2 of any columns whose name is in the `union` of the two `set`s and is not `<: Number`.
4. df = df[complete_cases(df2),:]`
as the very first line of the `DataFrame` recipe. That will drop all cases where the non-numeric argument is NA.

I am in favour of handling Strings explicitly as "" as mentioned above

tbreloff:
If the code works for non-null columns, then :+1: .  We can make it more robust in the future.

mkborregaard:
Should do:
```julia
using StatPlots, DataFrames
df = DataFrame(a = 1:10, b = randn(10), c = rand(1:2,10))
df[:a][3] = Nullable();df[:c][5] = Nullable()
p1 = plot(df, :a, [:b :c])
p2 = plot(df, :a, :b, group = :c)
plot(p1, p2)
```
skaermbillede 2017-01-19 kl 23 14 37

mkborregaard:
Any opinion on how to deal with the non-numerics as I outlined above?

tbreloff:
I'd need to think on it more.  I could see the argument that a null-value should just be "another discrete category".  For example if you wanted a bar plot representing a count of all values, then Nullable() would just be one of the values.

mkborregaard:
Yes, that is how R plotting handles NAs in non-numerics (e.g. bar plots). That would be achieved, I think, by converting to "" for Strings and Symbol() for Symbol. We should have good time to think on it more, but let me know if you reach a preference.

piever:
If I understand correctly there are two possible behvaviours: if some quantity is missing, either skip the datapoint, or assign it to a new discrete category. In my opinion, conceptually the criterion to decide which to do is not subtype of Number vs not subtype of Number, but rather NullableArray (skip datapoint) vs NullableCategoricalArray (create new category). The only thing to decide then would be how's the value of NULL printed with the new categories. Maybe one could even go so far as using `x.refs` (for `x::NullableCategoricalArray`) to actually build the plot and then replace those values on the plot with the corresponding value in `x.pool` if `refs>0` or with "NULL" if `refs == 0`

mkborregaard:
If I understand completely, I don't think I agree - all non-Numbers are treated as Categorical in Statplots (see also here https://github.com/JuliaPlots/StatPlots.jl/pull/36#pullrequestreview-18278479). Whether to drop, error or create a new category on categoricals is worth discussing with @tbreloff – I am personally in favour of just replacing with "" for String and Symbol() for Symbol.

piever:
What I meant goes actually in the other direction: I wanted to say that a categorical array of a subtype of Numbers (i.e. categorical array of integers or booleans) should be treated the same way as Strings and Symbols (which is what I do in `groupapply`). I agree with you that non-Numbers should be treated as categorical no matter what.

mkborregaard:
Oh, right, then I agree :-)

PR #35, opened 2017-01-20
Adds support for Nulls in String and Symbol DataFrame columns

mkborregaard:
Do not merge


mkborregaard:
Whether this is desired behaviour depends on whether the dataframe column is used for group, or is e.g. a column of markershape symbols. In the group case, it will treat Nulls as just another group instead of failing. It is also desirable e.g. for series_annotations.
I think the passing of explicit symbols in e.g. markershape (which would fail on nulls) is perhaps rare.

PR #36, opened 2017-01-21
Adapt groupapply to new DataFrames

piever:
Here are the changes for `groupapply` to work with the new DataFrames design. I simply check what columns will be used to produce the plot, exclude the datapoints where there's missing data in one of those columns, convert to a dataframe of arrays and then proceed as usual with the analysis and plotting.

### DIFF
--- a/src/groupederror.jl
+++ b/src/groupederror.jl
@@ -1,15 +1,14 @@
 ##### List of functions to analyze data
 
 function extend_axis(df::AbstractDataFrame, xlabel, ylabel, xaxis, val)
-    aux = DataFrame()
-    aux[xlabel] = xaxis
+    aux = DataFrame(Any[xaxis],[xlabel])
     extended = join(aux, df, on = xlabel, kind = :left)
     sort!(extended, cols = [xlabel])
     return convert(Array, extended[ylabel], val)
 end
 
 function extend_axis(xsmall, ysmall, xaxis, val)
-    df = DataFrame(x = xsmall, y = ysmall)
+    df = DataFrame(Any[xsmall,ysmall], [:x,:y])
     return extend_axis(df, :x, :y, xaxis, val)
 end
 
@@ -202,14 +201,21 @@ Output is a `GroupedError` with error computed according to the keyword `compute
 It can be plotted using `plot(g::GroupedError)`
 Seriestype can be specified to be `:path`, `:scatter` or `:bar`
 """
-function groupapply(f::Function, df, args...;
+function groupapply(f::Function, dfn, args...;
                     axis_type = :auto, compute_error = :none, group = [],
                     summarize = (get_symbol(compute_error) == :bootstrap) ? (mean, std) : (mean, sem),
                     compute_axis = :separate,
                     kwargs...)
     if !(axis_type in [:discrete, :continuous])
-        axis_type = (typeof(df[args[1]])<:PooledDataArray) ? :discrete : :continuous
+        axis_type = (typeof(dfn[args[1]])<:Union{CategoricalArray,NullableCategoricalArray}) ?
+                    :discrete : :continuous
     end
+    # Convert to dataframe with only
+    rel_cols = filter(t -> (t in names(dfn)), union([args...], compute_error, group))
+    df = dfn[:, rel_cols]
+    df = df[complete_cases(df),:]
+    df = DataFrame(convert.([Array],getindex.([df],rel_cols)) , rel_cols)
+
     if (axis_type == :continuous) & !(eltype(df[args[1]])<:Real)
         warn("Changing to discrete axis, x values are not real numbers!")
         axis_type = :discrete
--- a/src/groupederror.jl
+++ b/src/groupederror.jl
@@ -189,10 +189,13 @@ function get_groupederror(trend,variation, f, df::AbstractDataFrame, axis_type, 
     return get_groupederror(trend,variation, f, df::AbstractDataFrame, xvalues, ce, args...; kwargs...)
 end
 
+vectorify(v::AbstractArray) = v
+vectorify(v) = [v]
+
 """
 
     groupapply(f::Function, df, args...;
-                axis_type = :auto, compute_error = :none, group = [],
+                axis_type = :auto, compute_error = :none, group = Array{Symbol,1}(0),
                 summarize = (get_symbol(compute_error) == :bootstrap) ? (mean, std) : (mean, sem),
                 kwargs...)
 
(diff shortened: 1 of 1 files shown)


src/groupederror.jl
```diff
@@ -202,14 +204,24 @@ Output is a `GroupedError` with error computed according to the keyword `compute
...
                     compute_axis = :separate,
                     kwargs...)
+
     if !(axis_type in [:discrete, :continuous])
-        axis_type = (typeof(df[args[1]])<:PooledDataArray) ? :discrete : :continuous
+        axis_type = (typeof(dfn[args[1]])<:Union{CategoricalArray,NullableCategoricalArray}) ?
+                    :discrete : :continuous
     end
+
+    # Convert to dataframe with only relevant columns and no missing data
```
mkborregaard: So far, we have not yet to decided what to do when the DataFrame has missing values. I think it is better to wait with this and error on Nulls, to have consistency across the repo. Could you take this out, I will merge the rest and we can put this bit in an un-merged PR.

mkborregaard:
Great work! I have a single comment on not at this point to automatically remove Nulls.

src/groupederror.jl
```diff
@@ -190,10 +189,13 @@ function get_groupederror(trend,variation, f, df::AbstractDataFrame, axis_type,
...
 end
 
+vectorify(v::AbstractArray) = v
+vectorify(v) = [v]
+
 """
 
     groupapply(f::Function, df, args...;
-                axis_type = :auto, compute_error = :none, group = [],
+                axis_type = :auto, compute_error = :none, group = Array{Symbol,1}(0),
```
nalimilan: `Symbol[]` is shorter (same below).

src/groupederror.jl
```diff
@@ -202,14 +204,24 @@ Output is a `GroupedError` with error computed according to the keyword `compute
...
-                    axis_type = :auto, compute_error = :none, group = [],
+function groupapply(f::Function, dfn, args...;
+                    axis_type = :auto, compute_error = :none, group = Array{Symbol,1}(0),
                     summarize = (get_symbol(compute_error) == :bootstrap) ? (mean, std) : (mean, sem),
                     compute_axis = :separate,
                     kwargs...)
+
     if !(axis_type in [:discrete, :continuous])
-        axis_type = (typeof(df[args[1]])<:PooledDataArray) ? :discrete : :continuous
+        axis_type = (typeof(dfn[args[1]])<:Union{CategoricalArray,NullableCategoricalArray}) ?
```
nalimilan: Shouldn't `Array{String}` also be considered as categorical?

src/groupederror.jl
```diff
@@ -202,14 +204,24 @@ Output is a `GroupedError` with error computed according to the keyword `compute
...
+                    :discrete : :continuous
     end
+
+    # Convert to dataframe with only relevant columns and no missing data
+    err_col = (isa(compute_error, Tuple) && (compute_error[1] == :across)) ?
+                vectorify(compute_error[2]) : Array{Symbol,1}(0)
+    rel_cols = filter(t -> (t in names(dfn)), union([args...], err_col, vectorify(group)))
+    df = dfn[complete_cases(dfn[:, rel_cols]),rel_cols]
+    df = DataFrame(convert.([Array],getindex.([df],rel_cols)) , rel_cols)
```
nalimilan: Wouldn't `Array.(...)` work here? Can't you also do `df[rel_cols]` directly?

src/groupederror.jl
mkborregaard: Yes, I just had the same thought in response to the comment here: https://github.com/JuliaPlots/StatPlots.jl/pull/34#issuecomment-274056998
In StatPlots, both `Array{String}` and `Array{Symbol}` are automatically treated as categorical variables.

src/groupederror.jl
piever: `df[rel_cols]` won't work, because you need an array of columns for the dot syntax to work, whereas `df[rel_cols]` returns a dataframe. Then the array of column becomes an array of arrays, which is the only way to get a dataframe without nullable arrays AFAIK. `Array.(...)` is much more legible than what I have and it does just the same, I'll change to that.

src/groupederror.jl
piever: I totally agree, that's done [here](https://github.com/JuliaPlots/StatPlots.jl/blob/master/src/groupederror.jl#L213): if there's an Array (not Categorical) of not numbers (e.g. strings or symbols) and the user didn't set `axis_type = :discrete` I give a warning and convert do discrete axis (if this is standard policy in StatPlots maybe I can remove the warning). Now the two checks (is array categorical and is `eltype<:Real`) got separated because I thought it's easier to check for the eltype after converting to normal arrays, but it's easier to check whether a column is categorical before converting.

nalimilan: I think it makes more sense to treat as categorical without a warning when the user didn't expliciltly request a continuous axis.

src/groupederror.jl
nalimilan: Probably more readable: `DataFrame([Array(df[col]) for col in rel_cols], rel_cols)` or `DataFrame(; (col=>Array(df[col]) for col in rel_col)...)`.

src/groupederror.jl
```diff
@@ -1,15 +1,14 @@
...
-    aux[xlabel] = xaxis
+    aux = DataFrame(Any[xaxis],[xlabel])
     extended = join(aux, df, on = xlabel, kind = :left)
     sort!(extended, cols = [xlabel])
     return convert(Array, extended[ylabel], val)
 end
 
 function extend_axis(xsmall, ysmall, xaxis, val)
-    df = DataFrame(x = xsmall, y = ysmall)
+    df = DataFrame(Any[xsmall,ysmall], [:x,:y])
```
nalimilan: Do you really need to use this special constructor here?

src/groupederror.jl
mkborregaard: yes

src/groupederror.jl
piever: I think I do: if I don't use it, it may happen that I try and join a dataframe of nullablearrays with a dataframe of normal arrays, which gives an error. The approach of `groupapply` is to convert what the user gave to a df of only arrays and then only work with that type of dataframes.

nalimilan: Can you file a bug about this in DataFrames?

piever: Sure!

piever: Thanks for taking care of the bug. Still, I'll probably keep this somewhat uglier version that works no matter what until NullableArrays stabilizes, and then I'll go over the code again to see if it can be made more elegant.

mkborregaard: That is OK

piever:
I should've addressed all of the concerns. Changing from this to excluding rows with missing data is actually one line so we can change it back anytime very easily.

mkborregaard:
Alright, I will merge it tomorrow unless @nalimilan or @tbreloff have further comments here. Thanks again for the great work!

tbreloff:
No comments (though I admit to not looking at it closely)

mkborregaard:
Great!

PR #46, opened 2017-03-10
ignores symbols that are not dataframe names

mkborregaard:


mkborregaard:
This is a necessary bugfix.

PR #50, opened 2017-05-01
Update for RecipesBase, and Plots.jl cycle change

ChrisRackauckas:


src/StatPlots.jl
```diff
@@ -2,8 +2,8 @@
 module StatPlots
 
 using Reexport
-@reexport using Plots
-import Plots: Plot, cycle
+using RecipesBase, Plots
```
mkborregaard: Should probably rename the PR

mkborregaard: And why do you need to import RecipesBase?

ChrisRackauckas: >And why do you need to import RecipesBase?

Because I'm dumb. I thought I backtracked everything before squashing. Looks like I missed one of the changes.

PR #52, opened 2017-05-17
Added side parameter (:both - default, :right, :left) to violin plot recipe

sylvaticus:
Added a `side` parameter (default on :both, :right, :left) to the violin recipe so that users can plot asymmetric violin plots (keyword name and options as in [statsmodels ](http://www.statsmodels.org/0.6.1/generated/statsmodels.graphics.boxplots.violinplot.html)).

### DIFF
--- a/src/violin.jl
+++ b/src/violin.jl
@@ -15,7 +15,7 @@ function violin_coords(y; trim::Bool=false)
 end
 
 
-@recipe function f(::Type{Val{:violin}}, x, y, z; trim=true)
+@recipe function f(::Type{Val{:violin}}, x, y, z; trim=true, side=:both)
     xsegs, ysegs = Segments(), Segments()
     glabels = sort(collect(unique(x)))
     bw = d[:bar_width]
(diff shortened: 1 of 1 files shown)


mkborregaard:
Nice, can you post an example plot?

sylvaticus:
Doesn't make much sense (unless it is true that in Moscow are taller!), but here is a modification of the current example...

```
using DataFrames, Plots, StatPlots, RDatasets
singers = dataset("lattice","singer")
singers[:chorus] = "Scala"
singers2 = deepcopy(singers)
singers2[:Height] = singers2[:Height]+5
singers2[:chorus] = "Moscow"
singersTot = vcat(singers,singers2)
myPlot = violin(singers,:VoicePart,:Height, side=:right, marker=(0.2,:blue,stroke(0)), label="Scala")
violin!(singers2,:VoicePart,:Height, side=:left, marker=(0.2,:red,stroke(0)), label="Moscow")
```

violinplot

mkborregaard:
That looks great, can you update the readme.md with an example?

sylvaticus:
adding the output..

violinplot_asymmetric

PR #53, opened 2017-05-17
Added side parameter (:both - default, :right, :left) to violin plot recipe

sylvaticus:
Update of pull request https://github.com/JuliaPlots/StatPlots.jl/pull/52
(sorry, PkgDev.submit() has opened a new pull request instead of updating the old one..)


mkborregaard:
I liked the clearer (if silly) example from our conversation better - and could you add just a line stating that this is to demonstrate the :side keyword?

mkborregaard:
You can also assume the dataset is already loaded from the example above

mkborregaard:
Nice, thanks @sylvaticus !

PR #63, opened 2017-07-14
Groupapply

piever:
I've updated `groupapply` for Julia 0.6. The only thing not working was that `linspace(a,b,n)` doesn't return a `LinSpace` anymore, so I used a more general `Range` type for dispatch to work.
Now it's possible to compute the hazard rate of a column (useful for survival analysis) as a built-in analysis function.
I've changed the README from `using StatPlots` to `using StatPlots; using Plots` because it seems to me that now only the second option works and, if you don't know, it is not obvious to find out what went wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ groupedbar(rand(10,3), bar_position = :dodge, bar_width=0.7)
 
 
 ## groupapply for population analysis
-There is a groupapply function that splits the data across a keyword argument "group", then applies "summarize" to get average and variability of a given analysis (density, cumulative and local regression are supported so far, but one can also add their own function). To get average and variability there are 3 ways:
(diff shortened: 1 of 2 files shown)


mkborregaard:
Why should just `using StatPlots`not work?

mkborregaard:
Ah, that's https://github.com/JuliaPlots/StatPlots.jl/issues/64 . That is separate from this - if you remove that line from the readme I'll merge.

mkborregaard:
BTW, @piever I'd like to add you as a collaborator to StatPlots to give you push rights to the repo. That would make it easier to do minor updates to groupapply, and also allow you to handle PRs if you like - we could use an extra pair of hands, and I think you are one of the people who knows the repo well. Is that OK with you?

mkborregaard:
I've sent the invite.

piever:
Ok, I've took out the extra `using Plots` from the README. Thanks for the invitation, I'm happy to join the StatPlots team!

mkborregaard:
Great!

PR #70, opened 2017-08-14
Fix groupapply axis for density plots

piever:
Corrected x axis computation for kernel density plot (it used to have issues in case of distributions who grow steeply at their minimum, e.g. Gamma)

Before:

before

After:

after


mkborregaard:
Nice!

PR #71, opened 2017-08-16
Allow binning preprocessing

piever:
Added possibility to bin the x axis before analysing the data. It allows to use simpler analysis methods in case of continuous data:
- to get y as a function of x there was Loess regression, now it's also possible to do it by binning the x axis
- to get density of x, before there was kernel density estimation, now can be done by binning

Useful if Loess regression is too computationally expensive (often the case on less powerful machines) 
or simply if one wants to use simpler analysis that are easier to explain.

Example: density bar plot of binned data versus continuous estimation:

```julia
grp_error = groupapply(:density, school, :MAch; axis_type = :binned, nbins = 40, group = :Minrty)
plot(grp_error, line = :bar, color = ["orange" "turquoise"], legend = :topleft)

grp_error = groupapply(:density, school, :MAch; axis_type = :continuous, group = :Minrty)
plot!(grp_error, line = :path, color = ["orange" "turquoise"], label = "")
```

density

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ There is a groupapply function that splits the data across a keyword argument "g
 
 - `compute_error = :none`, where no error is computed or displayed and the analysis is carried out normally.
 
-The local regression uses [Loess.jl](https://github.com/JuliaStats/Loess.jl) and the density plot uses [KernelDensity.jl](https://github.com/JuliaStats/KernelDensity.jl). In case of categorical x variable, these function are computed by splitting the data across the x variable and then computing the density/average per bin. The choice of continuous or discrete axis can be forced via `axis_type = :continuous` or `axis_type = :discrete`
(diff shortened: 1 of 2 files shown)


mkborregaard:
You're really developing this functionality, I've gotta say!

piever:
It's a type of analysis we use a lot in the lab. I'm actually using the fact that this function can do most plots with more or less the same code structure to build a QML GUI to do analysis plots with a few clicks (tbh I'm not very fond of GUIs, but it has turned out to be very useful for my labmates).

A couple of questions:
- Travis build is not passing on OSX but for a silly reason: it seems it can't install plot https://travis-ci.org/JuliaPlots/StatPlots.jl/jobs/265470015#L140, should I just restart it? Does it happen often?
- After this PR is merged, is it possible to make a new StatPlots point release? `groupapply` only works well on StatPlots master on Julia 0.6

mkborregaard:
- That shouldn't happen often.
- Sure :-)

Issue #72, opened 2017-08-21
Discussion: syntax for plotting from a DataFrame


piever:
Issues with the current syntax for plotting from a DataFrame:
- It's implemented as a user recipe, which means it doesn't play well with grouping (see [#1033](https://github.com/JuliaPlots/Plots.jl/pull/1033))
- The user may be led to expecting that it is equivalent to typing the columns inside the DataFrame (which I believe it should be), but instead it is different in subtle ways, for example when trying to group by more than one column. This fails

```
using StatPlots, DataFrames, RDatasets
iris = RDatasets.dataset("datasets","iris")
iris[:LongSepal] = iris[:SepalLength] .> median(iris[:SepalLength])
scatter(iris, :PetalLength, :PetalWidth, group = collect(zip(:Species,:LongSepal)))
```

but it'd work with the columns:

`scatter(iris[:PetalLength], iris[:PetalWidth], group = collect(zip(iris[:Species],iris[:LongSepal])))`

- other recipes also fail because the transformation from symbol to columns happens to late (see #59 ):

`corrplot(iris, [:SepalLength :SepalWidth :PetalLength :PetalWidth])`

- it has to resort to heuristics to decide whether a symbol is a symbol or a column. For example 
`scatter(iris, :PetalLength, :PetalWidth, color = :blue)` plots in blue, unless there is a column in `iris` called `:blue`

- it has to resort to a runtime `eval` call

My personal preference would be for syntactically replacing symbols with columns.
Possible options:

- Adding a dispatch to `plot` if the first argument is a DataFrame: in that case all symbols and expression are replaced by the respective columns, using the same functions we use now. It wouldn't resolve the ambiguity symbol/column + calling `eval` but at least wouldn't mess with the recipe system

- Officialize the `@with` method from DataFramesMeta (a bit verbose but solves all of the above problems: `group` and `corrplot` work just fine and one can escape symbols to prevent them from becoming columns). Extra advantage: it can be used in blocks:

```
@with iris begin
    plt = plot(:SepalLength,:PetalWidth)
    plot!(:SepalLength,:PetalLength)
    display(plt)
end
```

- A slight modification of `@with` where each symbol `s` is not replaced by `df[s]`, but by 
`haskey(df, s) ? df[s] : s` so that the user wouldn't need to escape symbols so often
 
- Try and find a less verbose macro, though the implementation is not super straightforward. One bold possibility would be sth like:

```
macro plot(args...)
[...]

PR #74, opened 2017-08-24
added given macro to plot dataframes

piever:
A possible solution to #72 using metaprogramming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,12 @@ df = DataFrame(a = 1:10, b = 10*rand(10), c = 10 * rand(10))
 plot(df, :a, [:b :c])
 scatter(df, :a, :b, markersize = :(4 * log(:c + 0.1)))
 ```
-If you find an operation not supported by DataFrames, please open an issue. An alternative approach to the `StatPlots` syntax is to use the [DataFramesMeta](https://github.com/JuliaStats/DataFramesMeta.jl) macro `@with`. Symbols not referring to DataFrame columns must be escaped by `^()` e.g.
+If you find an operation not supported by DataFrames, please open an issue. An alternative approach to the above syntax is to use the macro `@given`. Symbols not referring to DataFrame columns must be escaped by `^()` wherever this could cause ambiguity e.g.
 ```julia
 using DataFramesMeta
-@with(df, plot(:a, [:b :c], colour = ^([:red :blue])))
+@given df plot(:a, [:b :c], colour = [:red :blue])
+df[:red] = rand(10)
+@given df plot(:a, [:b :c], colour = ^([:red :blue]))
 ```
 ---
 
--- a/README.md
+++ b/README.md
@@ -25,19 +25,23 @@ using StatPlots
 gr(size=(400,300))
 ```
 
-The `DataFrames` support allows passing `DataFrame` columns as symbols. Operations on DataFrame column can be specified using quoted expressions, e.g.
+`DataFrames` are supported thanks to the macro `@df` which allows passing `DataFrame` columns as symbols. Those columns can then be manipulated inside the `plot` call, like normal `Arrays`:
 ```julia
 using DataFrames
 df = DataFrame(a = 1:10, b = 10*rand(10), c = 10 * rand(10))
-plot(df, :a, [:b :c])
-scatter(df, :a, :b, markersize = :(4 * log(:c + 0.1)))
+@df df plot(:a, [:b :c], colour = [:red :blue])
+@df df scatter(:a, :b, markersize = 4 * log.(:c + 0.1))
 ```
-If you find an operation not supported by DataFrames, please open an issue. An alternative approach to the above syntax is to use the macro `@given`. Symbols not referring to DataFrame columns must be escaped by `^()` wherever this could cause ambiguity e.g.
+
+In case of ambiguity, symbols not referring to `DataFrame` columns must be escaped by `^()`:
 ```julia
-using DataFramesMeta
-@given df plot(:a, [:b :c], colour = [:red :blue])
 df[:red] = rand(10)
-@given df plot(:a, [:b :c], colour = ^([:red :blue]))
+@df df plot(:a, [:b :c], colour = ^([:red :blue]))
+```
+
+The old syntax, passing the `DataFrame` as the first argument to the `plot` call is still supported, but has several limitations (the most important being incompatibility with user recipes):
+```julia
+plot(df, :a, [:b :c], colour = [:red :blue])
 ```
 ---
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ iris = dataset("datasets","iris")
 @df iris marginalhist(:PetalLength, :PetalWidth)
 ```
 
-![](https://cloud.githubusercontent.com/assets/933338/19213780/a82e34a6-8d42-11e6-8846-80c9f4c48b9c.png)
+![marginalhist](https://user-images.githubusercontent.com/6333339/29869938-fbe08d02-8d7c-11e7-9409-ca47ee3aaf35.png)
 
 ---
 
@@ -67,7 +67,7 @@ or also:
 @df iris corrplot(cols(1:4), grid = false)
 ```
 
-![corrplot](https://user-images.githubusercontent.com/6333339/29830744-fd1fe282-8cda-11e7-86aa-ff5e8033a693.png)
+![corrplot](https://user-images.githubusercontent.com/6333339/29870023-7b07b010-8d7d-11e7-901c-3ef9a6af78bb.png)
 
 
 A correlation plot may also be produced from a matrix:
@@ -105,7 +105,7 @@ singers = RDatasets.dataset("lattice","singer")
 @df singers boxplot!(:VoicePart,:Height,marker=(0.3,:orange,stroke(2)))
 ```
 
-![](https://juliaplots.github.io/examples/img/pyplot/pyplot_example_30.png)
+![violin](https://user-images.githubusercontent.com/6333339/29870077-b4242e32-8d7d-11e7-9b18-40a57360936d.png)
 
 Asymmetric violin plots can be created using the `side` keyword (`:both` - default,`:right` or `:left`), e.g.:
 
@@ -116,7 +116,7 @@ singers_moscow[:Height] = singers_moscow[:Height]+5
 @df singers_moscow violin!(:VoicePart,:Height, side=:left, marker=(0.2,:red,stroke(0)), label="Moscow")
 ```
 
-![](https://cloud.githubusercontent.com/assets/2077159/26156938/22ccf0d4-3b18-11e7-9f34-555005437e6c.png)
+![2violin](https://user-images.githubusercontent.com/6333339/29870110-d90ed468-8d7d-11e7-8ebb-008323dff8b8.png)
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ plot(df, :a, [:b :c], colour = [:red :blue])
 ```julia
 using RDatasets
 iris = dataset("datasets","iris")
-marginalhist(iris, :PetalLength, :PetalWidth)
+@df iris marginalhist(:PetalLength, :PetalWidth)
 ```
 
 ![](https://cloud.githubusercontent.com/assets/933338/19213780/a82e34a6-8d42-11e6-8846-80c9f4c48b9c.png)
@@ -59,6 +59,17 @@ marginalhist(iris, :PetalLength, :PetalWidth)
 
 ## corrplot and cornerplot
 
+```julia
+@df iris corrplot([:SepalLength :SepalWidth :PetalLength :PetalWidth])
+```
+or also:
+```julia
+@df iris corrplot(cols(1:4))
+```
+
+![iris](https://user-images.githubusercontent.com/6333339/29792101-3daa6cfe-8c37-11e7-97e2-d1763d3b95c3.png)
+
+A correlation plot may also be produced from a matrix:
 
 ```julia
 M = randn(1000,4)
@@ -89,8 +100,10 @@ cornerplot(M, compact=true)
 ```julia
 import RDatasets
 singers = RDatasets.dataset("lattice","singer")
-violin(singers,:VoicePart,:Height,marker=(0.2,:blue,stroke(0)))
-boxplot!(singers,:VoicePart,:Height,marker=(0.3,:orange,stroke(2)))
+@df singers begin
+    violin(:VoicePart,:Height,marker=(0.2,:blue,stroke(0)))
+    boxplot!(:VoicePart,:Height,marker=(0.3,:orange,stroke(2)))
+end
 ```
 
 ![](https://juliaplots.github.io/examples/img/pyplot/pyplot_example_30.png)
@@ -100,8 +113,8 @@ Asymmetric violin plots can be created using the `side` keyword (`:both` - defau
 ```julia
 singers_moscow = deepcopy(singers)
 singers_moscow[:Height] = singers_moscow[:Height]+5
-myPlot = violin(singers,:VoicePart,:Height, side=:right, marker=(0.2,:blue,stroke(0)), label="Scala")
-violin!(singers_moscow,:VoicePart,:Height, side=:left, marker=(0.2,:red,stroke(0)), label="Moscow")
+@df singers violin(:VoicePart,:Height, side=:right, marker=(0.2,:blue,stroke(0)), label="Scala")
+@df singers_moscow violin!(:VoicePart,:Height, side=:left, marker=(0.2,:red,stroke(0)), label="Moscow")
 ```
 
 ![](https://cloud.githubusercontent.com/assets/2077159/26156938/22ccf0d4-3b18-11e7-9f34-555005437e6c.png)
--- a/README.md
+++ b/README.md
@@ -60,14 +60,15 @@ iris = dataset("datasets","iris")
 ## corrplot and cornerplot
 
 ```julia
-@df iris corrplot([:SepalLength :SepalWidth :PetalLength :PetalWidth])
+@df iris corrplot([:SepalLength :SepalWidth :PetalLength :PetalWidth], grid = false)
 ```
 or also:
 ```julia
-@df iris corrplot(cols(1:4))
+@df iris corrplot(cols(1:4), grid = false)
 ```
 
-![iris](https://user-images.githubusercontent.com/6333339/29792101-3daa6cfe-8c37-11e7-97e2-d1763d3b95c3.png)
+![corrplot](https://user-images.githubusercontent.com/6333339/29830744-fd1fe282-8cda-11e7-86aa-ff5e8033a693.png)
+
 
 A correlation plot may also be produced from a matrix:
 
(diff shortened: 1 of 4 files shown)


src/given.jl
```diff
@@ -0,0 +1,39 @@
...
+select_column(df, s) = haskey(df, s) ? convert_column(df[s]) : s
+
+
+convert_column(col) = col
+
+function convert_column(col::AbstractDataArray{T}) where T
+    try
+        convert(Array, col)
+    catch
+        error("Missing data of type $T is not supported")
```
mkborregaard: A possible, more featured, alternative, would be to tally NAs in undefined columns and then remove those from all input columns. We don't need that for now though.

src/given.jl
```diff
@@ -0,0 +1,39 @@
...
+    end
+end
+
+convert_column(col::AbstractDataArray{String})  = convert(Array, col, "")
+convert_column(col::AbstractDataArray{Symbol})  = convert(Array, col, Symbol())
+convert_column(col::AbstractDataArray{Float64}) = convert(Array, col, NaN)
```
mkborregaard: I think this could be `convert_column(col::AbstractDataArray{<:Real}) = convert(Array, convert(DataArray{Float64},col), NaN)` for consistency with other number types.

README.md
```diff
@@ -32,10 +32,12 @@ df = DataFrame(a = 1:10, b = 10*rand(10), c = 10 * rand(10))
 plot(df, :a, [:b :c])
 scatter(df, :a, :b, markersize = :(4 * log(:c + 0.1)))
 ```
-If you find an operation not supported by DataFrames, please open an issue. An alternative approach to the `StatPlots` syntax is to use the [DataFramesMeta](https://github.com/JuliaStats/DataFramesMeta.jl) macro `@with`. Symbols not referring to DataFrame columns must be escaped by `^()` e.g.
```
mkborregaard: I think we should highlight this as the preferred approach, and note the current approach as a 'legacy' approach further down. No need to break the functionality but it is less powerful.

src/given.jl
```diff
@@ -0,0 +1,39 @@
+"""
+    `@given d x`
```
mkborregaard: an alternative could be `@use`? Need to think about names.

src/given.jl
```diff
@@ -0,0 +1,39 @@
...
+
+function _given(d, x::Expr)
+    (x.head == :quote) && return :(StatPlots.select_column($d, $x))
+    (x.head == :call) && x.args[1] == :^ && length(x.args) == 2 && return x.args[2]
+    return Expr(x.head, _given.(d, x.args)...)
+end
+
+
+select_column(df, s) = haskey(df, s) ? convert_column(df[s]) : s
+
```
mkborregaard: single-spaced

src/given.jl
```diff
@@ -0,0 +1,39 @@
...
+function convert_column(col::AbstractDataArray{T}) where T
+    try
+        convert(Array, col)
+    catch
+        error("Missing data of type $T is not supported")
+    end
+end
+
+convert_column(col::AbstractDataArray{String})  = convert(Array, col, "")
```
mkborregaard: AbstractString

src/StatPlots.jl
```diff
@@ -17,7 +17,9 @@ import Loess
 
 export groupapply
```
mkborregaard: could the groupapply functions be rewritten to use this format instead?

piever: Well, at the very least the macro can certainly translate:
`@given df groupapply(func, x,....)` to `groupapply(func, df, x,....)`
But potentially it could be done in a smarter way so that it would also add support a general table and not only DataFrames and allow for some data manipulation (i.e. group by sth other than a column, for example `:SepalLength .> 3`).
One thing that worries me is the order of the arguments. As of now there are several ways to call it (example after the arrow):

```julia
groupapplay(func, df, x) ~> groupapplay(:density, iris, :SepalLength)
groupapplay(func, df, x, y) ~> groupapplay(:locreg, iris, :SepalLength, :SepalWidth)
```

And 2 convenience methods for plotting `y` as a fct of `x`:

```julia
groupapplay(y, df, x) == groupapplay(:locreg, df, x, y)
groupapplay(df, x, y) == groupapplay(:locreg, df, x, y)
```
The rational behind the first convenience method is that, when called in the standard way, the first variable represents what's on the `y` axis. Without `df` in the function call, these two methods would be ambiguous and we'd need to decide which one to keep.

mkborregaard: Scrap `groupapplay(y, df, x) == groupapplay(:locreg, df, x, y)` IMHO.

mkborregaard: It'd be elegant to reinterpret groupapply in the light of `given`. Intuitively I'd think it could be simplified a lot, but I may be wrong.

src/given.jl
piever: Regarding the name, I think that brevity is essential as this will be typed pretty often. I like  `@in` to denote that you are working within a given DataFrame, but `@use` also makes sense.

src/StatPlots.jl
piever: Agreed on both.
I also think `groupapply` can benefit from this (and also from the adjustments to the `group` keyword in [#1033](https://github.com/JuliaPlots/Plots.jl/pull/1033) ) but I need to think carefully about it. It probably deserves its own issue.

In terms of syntax, one idea would be to also allow:
```julia
@given df groupapply(...) plot(; kwargs....)
```
to mean the same as
```julia
plot(@given df groupapply(...); kwargs....)
```
But I guess we should open a separate issue to discuss changes to `groupapply`

mkborregaard:
@tkelman we're considering adding this PR or something similar that adds a macro (called `@given`, `@in` or `@use`, which implements a behaviour similar to `@with` for DataFrames (and possibly a future abstract table type), i.e. it looks in the args and replaces symbols with DataFrame columns, removes missing values etc. Would you accept that in a release of StatPlots or do you see clashes with other functionality in the ecosystem? 
I should add that it appears to be the best way to support DataFrame plotting in this package.

tkelman:
Worth maybe searching through if that's a common export by many other packages, but that seems fine? I'd recommend having people try it out on master for a while with various other packages and see if "both ... and StatPlots export ..." warnings are something a lot of people see. They can always switch from `using` to `import` but not everyone likes doing that.

mkborregaard:
Perfect, thanks a lot!

src/StatPlots.jl
mkborregaard: Good idea with a separate issue.

src/given.jl
mkborregaard: or `@df`?

piever: Also one of my labmates and I came to `@df` independently, I think it's a winner.

piever:
Addressed all feedback (I think) and renamed the macro to `@df`. I'd suggest maybe merging this soonish to see if there are any complaints and then move on with `groupapply` changes (it needs a few, I'd like to make it compatible with the new `@df` syntax as well as the new grouping in Plots)

mkborregaard:
Before merging we should update the entire readme (up until groupapply) to use the new syntax too - we could keep the old call commented out as an alternative.

piever:
> Nice, thanks for doing this! Let me just give it a spin tomorrow before we decide to merge?

Sure, I can update the README in the meantime.

piever:
And another thing that deserves consideration is how to add the axis labels automatically in the macro framework. It's not completely trivial (and it was not trivial even with the old syntax: `scatter(iris, :SepalLength)` puts the label on the x axis when it would refer to the y axis).

To get the labels, now the user would have to type:
`@df iris scatter(:SepalLength, :SepalWidth, xlabel = ^(:SepalLength), ylabel = ^(:SepalWidth))
`
which is really not ideal and should be automatised by the macro.
My suggestion is to add a kw argument to the plot call, called default_label where the macro would write all the relevant information (namely the list of args as it found it). For example this
`@df iris scatter(:SepalLength, :SepalWidth)`
would actually compile to:
`scatter(iris[:SepalLength], iris[:SepalWidth], default_label = (:SepalLength, :SepalWidth))`
and then plot would know that when there are 2 `default_label` one is for `xlabel` and one for `ylabel` unless they are specified by the user etcetera...
I'd rather do it with a separate kw because the translation is not obvious, as sometimes it refers to xlabel, ylabel and zlabel and sometimes it refers to label directly, such as in:

`corrplot(M, label = ["x$i" for i=1:4])`

But for this we'd first need to update Plots.jl so that if given a `default_label = (args...)`, it knows how to use it.

src/df.jl
```diff
@@ -0,0 +1,37 @@
...
+`NA` values are replaced with `NaN` for columns of `Float64` and `""` or `Symbol()`
+for strings and symbols respectively.
+"""
+macro df(d, x)
+    esc(_df(d,x))
+end
+
+_df(d, x) = x
+
+function _df(d, x::Expr)
```
mkborregaard: Perhaps here we could have some convenience functions here that are useful for plotting with DataFrames. Like allowing the symbol `:nrow` to be replaced by `size(x,1)`. 
Would it also be nice to be able to specify columns by number, like `cols(1:4)` and get `reshape(names(d)[x],1,:)`?

mkborregaard:
Why not just have the macro check if any of the kw `xlabel`, `ylabel`, `zlabel` or `label` appear in the call, and if they don't them put in the default labels?

piever:
> Why not just have the macro check if any of the kw xlabel, ylabel, zlabel or label (or aliases) appear in the call, and if they don't then put in the default labels?

I was afraid the presence of aliases/magic attributes could make this a nightmare, for example imagine the user gives:

`plot(y, xaxis = ("my label", (0,10), 0:0.5:10, :log, :flip, font(20, "Courier")))`

from the plots docs with magic attributes. The macro will never find out that `xlabel` is already assigned. I think it's best if the macro passes on the information about the symbols it replaced, as they may be used later on, in a kw with that purpose (I like `default_labels`, but maybe `arg_names` makes more sense) and Plots.jl will take care of using them.

src/df.jl
piever: That's an interesting idea, especially for `cols(1:4)` (for let's say a `corrplot`). Two questions:
- what's an example use of `size(x,1)` in a plot call?
- should we go for something like: `df` gets replaced with the `DataFrame`? In that case maybe `:nrow` can be replaced by `nrow(df)` (which would work for free) and we'd get access to all the `DataFrame` functions (though, to be honest, I'm not sure how many would actually be useful)

mkborregaard: - `marker_z = 1:size(x,1)`
- No, I'm not sure that would be necessary or clean

piever: Agreed on point 2. About point 1, there's the issue that `marker_z = 1:(:nrow)` is not the prettiest syntax, plus we create even more ambiguity with the symbols: what if a column is called `:nrow`? Maybe simply `nrow` and `ncol` are used (normal function names act on the DataFrame and symbols are for columns). Now `cols(1:ncol)` does the full `corrplot`.

daschw: Are these extra features compatible to the plan to support a variety of data table tapes?
If not: Do we even still pursue this plan?

mkborregaard: My take - we must pursue that long-term, but it seems early to me to start second-guessing what the interface will be.

mkborregaard: @piever you have a point on nrow and friends, let's keep that out of it. When you say "now cols(1:ncol) does" do you mean that you've already added the functionality but not pushed it?

mkborregaard:
Good point on the magic arguments but I'm not keen on adding an extra kw. I need to think about this more - perhaps we can somehow plug this into the recipe system.

src/df.jl
piever: `cols` yes, `nrow` it depends (or at least it would require some thinking, depending on which plan we choose). I think the plan to support multiple data source has at least two possible implementations:
- `@df` converts the first argument to a DataFrame (using IterableTables) and then proceeds as usual: anything that'd work on a DataFrame keeps working. It could be inefficient to convert the whole thing though (plotting a big DataFrame is rare, but I often use `groupapply` on pretty large data)
- the only thing that is DataFrame specific, gets done with IterableTables: `df[col]` is changed to `[getindex(s, col) for s in df]` (I need to see how it'd work with missing data)

Getting plan 2 to work with `groupapply` may require some thinking, so I thought it's better to postpone this decision, until adjusting `groupapply` to the new syntax.

mkborregaard: I like 2.

piever: > When you say "now cols(1:ncol) does

It's more like wishful thinking and poor phrasing... I meant with the hypothetical implementations. Still, if we decide that `nrow` and `ncol` are left out for the time being, implementing `cols` is pretty easy, I should be able to add it and push today.

piever:
Maybe I was too pessimistic. If both explicit argument and magic argument are given, only the magic argument is taken into account. Your solution is probably best then

`plot(rand(10), xaxis = ("my label"), xguide = "my new label")`

newplot-7

piever:
Actually it's even easier than that, as the rightmost keyword takes precedence, so it's enough to add the automatic keyword in the leftmost position to give it low priority.

piever:
I think it has all the feature it needs (if we agree to wait for `:nrow` and `:ncol`). It only has one bug on `marginalhist` but I think the bug is on `marginalhist` side, I've reported it here: #79. Feel free to play a bit with it and let me know if you have more improvements, example usage is in the README.

mkborregaard:
No, this is because my suggestion of adding xlabel to the plot call wasn't fully thought through. Doing it this way will override any labels set in a recipe with `-->` (which is commonly used for labels). That's unfortunate, but I think it will only be an issue with user recipes. One way to deal with this might be to only change it if the the plotting command is one of `:scatter`, `:plot`, and perhaps the `:histogram` types?

mkborregaard:
Also, if you have 
```julia
df = DataFrame(a = randn(1000))
@df df plot(:a)
```
You'll get the marker wrongly on the x axis.

piever:
I'm also a bit unsure about adding `xlabel` (it feels hacky) and would overall prefer the extra keyword `argnames`, to allow more flexibility. 

Still, in this particular case, I really think `marginalhist` is at fault, because even this command:

`marginalhist(randn(1000), randn(1000), xlabel = :x , ylabel = :y)
`
gives what I believe is the wrong plot:

marginal

I think that [this line](https://github.com/JuliaPlots/StatPlots.jl/blob/master/src/marginalhist.jl#L25) was supposed to force both axis labels to be `""`, but only works on the x axis label.

In general, I can not think of many recipes that rely on the user not giving x and y labels.

As for the second point,

```julia
df = DataFrame(a = randn(1000))
@df df plot(:a)
```

gives the same as the old syntax:

`plot(df, :a)`

The big issue is that, if there is only one argument, it is really unclear where the label should go: here it should go on the y axis whereas on `@df df density(:a)` it should go on the x axis. I think it's hard to guess from the function call where it should go.

Possible solutions:

- Implement the extra keyword `argnames` where to store all the argument names and then use runtime information to figure out the best label placement
- recommend `@df plot(1:nrow, :a)` instead

mkborregaard:
> I really think marginalhist is at fault

Sure, there's also the problem that `guide` and `label` are aliases, but both are first-class keywords. IMHO we should make guide just an alias of label, that would necessitate shifting guide to label in all recipes.

> would overall prefer the extra keyword argnames

That's cleaner here , but how would it be implemented inside Plots? We already have the recipes system for overloading keywords. If you can come up with a really clean way of implementing `default_names` then maybe that's the way to go. But it's tricky, e.g. how does it handle `plot!`?

> it is really unclear where the label should go

By convention, it should go to the y axis, except for the case of the histogram-like methods (which are unique in plotting y on the x axis).

> I'm also a bit unsure about adding xlabel (it feels hacky)

It does. The nice thing about is that we don't have to embed the StatPlots functionality deep within Plots - everything could be self-contained within this macro. I think only doing it for a call of `plot` or `scatter` should be safe, and feature-wise comparable to what we had.

> recommend @df plot(1:nrow, :a) instead

No. That would be against the core philosophy of Plots.

> In general, I can not think of many recipes that rely on the user not giving x and y labels.

Most multi-panel ones. And e.g. andrewsplot in PlotRecipes.


I am really sorry for disagreeing on a number of points here, considering that you've done such great work on this. One possibility would be to remove the automatic labels for now, and merge the rest as is? And we can continue working on the labels.

piever:
> I am really sorry for disagreeing on a number of points here

You are raising several good points, I'm also not fully convinced by this design. I can probably keep the helper function to extract column names as it will be useful in the future, take out the labelling part and merge as is.

Of all the concerns, `plot!` is probably the most catastrophic: one certainly doesn't want to choose the labels carefully on the first plot call and then get all of them overridden accidentally.

I think that the `default_names` approach may be the only way forward for the automated labelling. It does require a bit of work on the Plots side, so it's better to postpone it. My idea is that [here](https://github.com/JuliaPlots/Plots.jl/blob/de32b02c759d26eff2706b467b084188d0997959/src/args.jl#L296) the axis guide default would not be `""` but rather it would look into the `default_names` and would choose it from there. This default value never overrides anything else, so it should be pretty safe. Maybe we should clean and merge this and then I can try to have branches here and in Plots in parallel to see if this idea works.

mkborregaard:
Good idea 👍

piever:
I've taken out the controversial labelling part (while keeping the unexported auxiliary function to collect the `default_names` from a plot call). I'm planning to merge later today and then proceed to:
- See if I can get a `default_names` kw to make sense in Plots.jl using `_axis_defaults[:guide]` (I think it's especially important for plots like a `corrplot` where it may be hard to remember what each column is)
- Figure out how to update `groupapply` (in a separate PR/issue)

Regarding 2, I think the consistent way to go is to write symbols in the `groupapply` call which would be then translated to arrays by the `@df` macro and then the `get_groupederror` function would take care of the splitting, averaging etcetera. I have a slightly off-topic question: at least at a first analysis, it would seem that the data manipulations inside `get_groupederror` are both easier to express and faster to execute using `IndexedTables` rather than `DataFrames` (these manipulations are invisible to the user, it is a matter of internally handling the data). Should that be the case, would it be possible to add an IndexedTables dependency or is it outside the scope of this package?

mkborregaard:
Sounds good. This seems good to merge when you feel it's ready.

WRT to the IndexedTables dep, I think in the long run we should support whatever infrastructure is needed to support tables, but it seems a little strange to support DataFrames outwardly and then use indexedTables on the inside. 

Do you think we could use the DataStreams.Sink type to accept input from any kind of table that implements the DataStreams.Source interface?

piever:
> WRT to the IndexedTables dep, I think in the long run we should support whatever infrastructure is needed to support tables, but it seems a little strange to support DataFrames outwardly and then use indexedTables on the inside.

It does, but the idea is that (modulo missing data issues which need to be clarified) with the tiny IterableTables dependency it's possible to extract columns of more or less anything, without adding a dependency on it:

> The package currently has support for the following data sources: DataFrames, DataStreams (including CSV, Feather, SQLite, ODBC), DataTables, IndexedTables, TimeSeries, Temporal, TypedTables, JuliaDB, DifferentialEquations (any DESolution), CSVFiles, ExcelFiles, FeatherFiles, StatFiles and any iterator who produces elements of type NamedTuple

```julia
julia> using IterableTables

julia> [i.Species for i in getiterator(iris)]
150-element Array{String,1}:
 "setosa"   
 "setosa"   
 "setosa"   
 "setosa"   
 "setosa"   
 "setosa"   
 "setosa"   
 "setosa"   
 "setosa"   
 "setosa"   
 "setosa"   
 ⋮          
 "virginica"
 "virginica"
 "virginica"
 "virginica"
 "virginica"
 "virginica"
 "virginica"
 "virginica"
 "virginica"
 "virginica"
 "virginica"

```

And the same code would work with more or less anything instead of `iris` (IndexedTable, CSV source, SQLite database, I'm investigating on Discourse if one can also do it with MySQL databases). We would actually not even need to depend on DataFrames.

What is different in the case of `groupapply` is that, once we collect the arrays with the data, there is a fair bit of data shuffling left to do the plot. My intuition is that the native DataFrames API is not the most efficient at doing that, in that IndexedTables can use a very smart indexing system to optimize this data shuffling.

mkborregaard:
If you think we could replace the DataFrames dep with and IterableTables dep and still support DataFrame inputs I'm all for it. But could we still support missing values? That's a real issue.

piever:
I'm pretty sure you can (modulo a bug in IterableTables with PooledDataArrays), I'll make a PR as soon as I have a completely fleshed out solution.

mkborregaard:
We could keep it for a transition period until the interface crystallizes more.

mkborregaard:
:tada:

PR #75, opened 2017-08-27
Added equal area histograms

mkborregaard:


mkborregaard:
@daschw , as you can see on the image, `linecolor` appears to be playing tricks, possibly after your change?
```julia
a = [randn(100); randn(100)+3; randn(100)/2+3]
ea_histogram(a, bins = :scott, fillalpha = 0.4)
```

skaermbillede 2017-08-27 kl 23 55 04

daschw:
https://github.com/JuliaPlots/Plots.jl/blob/master/src/args.jl#L83 would need to be extended by `:ea_histogram` for this check here https://github.com/JuliaPlots/Plots.jl/blob/master/src/args.jl#L1267, right?

daschw:
Alternatively, maybe we could add
```julia
const allHistogramTypes = [:histogram, :barhist, :barbins]
like_histogram(seriestype::Symbol) = seriestype in allHistogramTypes
```
in Plots and in StatPlots:
```julia
push!(Plots.allHistogramTypes, :ea_histogram)
```
?

mkborregaard:
Ah, that's it of course.

mkborregaard:
In fact I actually don't think the best thing is to make this a new series, because it clearly is a histogram, with uneven bins and density-normalized. One suggestion could be to add this to Plots.jl instead, and then call it with `normalize = :equalarea`? 
It is a bit tricky as it isn't strictly a normalization, but that would make it nice and compact with the Plots syntax.

daschw:
:+1: from my side for fitting nice with Plot syntax.
@oschulz?

mkborregaard:
I thought this was about your new linecolor change :flushed:

oschulz:
The problem with `normalize = :equalarea` is that it precludes applying a "real" normalization at the same time. Since the equal-area histogram is actually just a specific binning, it would be nice if it could somehow be expressed via the `bins` keyword. Maybe something like `bins = EqualArea()` and `bins = EqualArea(100)` to optionally specify the number of bins? `EqualArea` being a struct?

mkborregaard:
The idea with using that keyword is that you'd never use another normalisation - the ea_histogram is by definition a specific binning + the :density normalisation. So the keyword should be free.

oschulz:
I could see it being used together with a :pdf normalization, e.g. to compare datasets with a different number of samples.

mkborregaard:
Yes, that would be possible. We could scrap density and make pdf the only option.
Alternatively we keep it as a distinct seriestype. That would just preclude people from using this with stephist.

oschulz:
I like the ability of being able to use it with different `stephist`, etc. How about having both `normalize = :equalarea_density` and `normalize = :equalarea_pdf`?

mkborregaard:
no, too wordy. Let's keep it as an alternative seriestype for now.

mkborregaard:
That also makes the implementation a lot simpler, as I can simply stick with this recipe :-) I tried faffing around with the normalise options, and it was a lot less clean.

mkborregaard:
Now depends on https://github.com/JuliaPlots/Plots.jl/pull/1058

oschulz:
All right ... would still be nice to be able to use it with another drawing style, but ...

Maybe we need to overhaul the auto-binning interface at some point, to allow more complex things there.

mkborregaard:
No, I agree. I just don't want the interface to Plots to be very complex to allow the user to set all kinds of details. Plots is meant to be easy and intuitive. In my opinion, not being able to have density (rather than pdf) transformed versions of equal-area histograms was a very small price to pay for clarity. But then again, you could argue that ea_histogram is a seriestype along the same lines as histogram2d.

mkborregaard:
Note that the current implementation doesn't allow changing the normalization either, though.

oschulz:
It would be nice if normalization could be specified, but I guess the implementation can be changed to support that later on.

PR #77, opened 2017-08-29
Allow specifying :wand bins

mkborregaard:
Depends on https://github.com/JuliaPlots/Plots.jl/pull/1053

This allows specifying binning using the data-based criterion of Wand, M.P. (1997) Data-Based Choice of Histogram Bin Width. The American Statistician, 51, 59-64. This belongs to a different, more modern class of histogram implementations where the correct bin size number is not only selected based on sample size, but also based on the shape of the distribution. Essentially it creates a kernel density and then finds the bin positions that minimizes the discrepancy between the histogram and the density function.

The code is a port of the `dpih` and associated functions found here: https://github.com/cran/KernSmooth/tree/master/R - the original license is "Unlimited" and I have the author's (Wand himself) permission to use it.

The bins are different from what we otherwise use in Plots as the bin edges do not align nicely with integer multipla of 1, 2 and 5. I wrote to Wand to ask him, and he insisted that was the most statistically valid in his view.

Comparisons using a skewed mixture distribution with 60, 300, 3000 and 30000 samples:
```julia
using Distributions, Plots
m = MixtureModel([Normal(),Normal(3, 2//3), Normal(2, 1//3)])
samples = (rand(m,x) for x in [60, 300, 3000, 30000])

gr(size = (1200, 1000), legend = false)
hists = [histogram(x, title = y, bins = y) for x in samples for y in [:auto, :scott, :wand]];
plot(hists..., layout = (4,3))
```
skaermbillede 2017-08-29 kl 14 34 09

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
-julia 0.6-pre
+julia 0.6
 
 Reexport
-Plots 0.12.0
+Plots 0.12.4
 StatsBase
 Distributions
 DataFrames
--- a/src/StatPlots.jl
+++ b/src/StatPlots.jl
@@ -17,7 +17,6 @@ import Loess
 
 export groupapply
 export get_groupederror
-export wand_edges
 export @df
 
 include("df.jl")
--- a/src/StatPlots.jl
+++ b/src/StatPlots.jl
@@ -17,6 +17,7 @@ import Loess
 
 export groupapply
 export get_groupederror
+export wand_edges
 export @df
 
 include("df.jl")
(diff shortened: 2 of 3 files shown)


mkborregaard:
I've chosen to PR this to StatPlots rather than to Plots as there's quite a lot of code and it may only be interesting for those with statistical interests. I'm very open to discuss that design, though.

mkborregaard:
BTW, you've gotta love the tersity and conciseness of the Plots code needed to generate that 12-panel plot. 
And even more BTW, the samples should really have been taken as
```julia
using Distributions
m = MixtureModel([Normal(),Normal(3, 2//3), Normal(2, 1//3)])
smallest, small, medium, large = [rand(m,x) for x in [20, 100, 1000, 10000]]
```
but who cares.

PR #82, opened 2017-08-31
WIP: add compatibility to a general table type and remove DataFrames dependency

piever:
This is work in progress to pass from a DataFrames based implementation of `@df` to a IterableTables base implementation. The following commands now give the desired plots (should work for everything that is supported by IterableTables):

```julia
using IterableTables, DataFrames, IndexedTables, RDatasets, CSV, StatPlots
iris = dataset("datasets","iris")
writetable("/tmp/iris.csv", iris)
@df CSV.Source("/tmp/iris.csv") scatter(:SepalLength, :SepalWidth)
iris[:SepalLength][2] = NA
iris_table = IndexedTable(iris)
@df iris_table scatter(:SepalLength, :SepalWidth)
```

The implementation of the `@df` macro no longer requires a DataFrames dependency. It works in the following way: IterableTables provides a `getiterator` which, from a table, creates a row based iterator which spits out every row as a named tuple. What I do here is to replace every symbol in the plot call with a new variable. Then the macro generates 2 commands: one to assign to the new variables the respective columns (using the function `StatPlots.compute_all`) and the other to do the  plot:

```julia
julia> @macroexpand @df iris_table scatter(:SepalLength, :SepalWidth)
quote 
    (##SepalLength#730, ##SepalWidth#731) = StatPlots.compute_all(iris_table, :SepalLength, :SepalWidth)
    scatter(##SepalLength#730, ##SepalWidth#731)
end
```

The advantage of doing all variables at once (and knowing exactly which they are) is twofold:
- As they are all replaced together, we can see which rows have unsupported missing data and potentially remove them with a warning
- the `getiterator` has to be iterated only once, which is better especially in the case of CSV or remote databases

I still need to polish a bit and further optimize `StatPlots.compute_all` but I wanted to check if this is a direction we're happy to take.

### DIFF
--- a/src/StatPlots.jl
+++ b/src/StatPlots.jl
@@ -8,7 +8,8 @@ using StatsBase
 using Distributions
 using DataFrames
 using IterableTables
-using DataValues
+import DataValues: DataValue
+import TableTraits: column_types, column_names
 
 import KernelDensity
 import Loess
--- a/src/StatPlots.jl
+++ b/src/StatPlots.jl
@@ -7,6 +7,8 @@ import Plots: _cycle
 using StatsBase
 using Distributions
 using DataFrames
+using IterableTables
+using DataValues
 
 import KernelDensity
 import Loess
--- a/src/StatPlots.jl
+++ b/src/StatPlots.jl
@@ -7,9 +7,10 @@ import Plots: _cycle
 using StatsBase
 using Distributions
 using DataFrames
-using IterableTables
+import IterableTables
 import DataValues: DataValue
-import TableTraits: column_types, column_names
+import TableTraits: column_types, column_names, getiterator, isiterabletable
+import TableTraitsUtils: create_columns_from_iterabletable
 
 import KernelDensity
 import Loess
--- a/src/df.jl
+++ b/src/df.jl
@@ -55,15 +55,7 @@ function select_column(df, s)
 end
 
 convert_missing(el) = el
-
-function convert_missing(el::DataValue{T}) where T
-    try
-        get(el)
-    catch
-        error("Missing data of type $T is not supported")
-    end
-end
-
+convert_missing(el::DataValue{T}) where {T} = get(el, error("Missing data of type $T is not supported"))
 convert_missing(el::DataValue{<:AbstractString}) = get(el, "")
 convert_missing(el::DataValue{Symbol}) = get(el, Symbol())
 convert_missing(el::DataValue{<:Real}) = get(convert(DataValue{Float64}, el), NaN)
--- a/src/df.jl
+++ b/src/df.jl
@@ -60,18 +60,15 @@ end
 
 stringify(x) = filter(t -> t != ':', string(x))
 
-compute_name(df, i) = column_names(IterableTables.getiterator(df))[i]
+compute_name(df, i) = column_names(getiterator(df))[i]
 
 function compute_all(df, syms...)
-    iter = IterableTables.getiterator(df)
-    type_info = Dict(zip(column_names(iter), column_types(iter)))
-    cols = Tuple(isa(s, Integer) ? Array{column_types(iter)[s]}(0) :
-        s in column_names(iter) ? Array{type_info[s]}(0) : s for s in syms)
-    for i in iter
-        for ind in 1:length(syms)
-            isa(cols[ind], Symbol) || push!(cols[ind], getfield(i, syms[ind]))
-        end
-    end
+    isiterabletable(df) || error("Only iterable tables are supported")
+    iter = getiterator(df)
+    name2index = Dict(zip(column_names(iter), 1:length(column_names(iter))))
+    col_index = [isa(s, Integer) ? s : get(name2index, s, 0) for s in syms]
+    cols = convert(Array{Any}, collect(syms))
+    cols[col_index .!= 0] = create_columns_from_iterabletable(df, filter(t -> t != 0, col_index))[1]
     return Tuple(convert_missing.(t) for t in cols)
 end
 
--- a/src/df.jl
+++ b/src/df.jl
@@ -9,19 +9,31 @@ If you want to avoid replacing the symbol, escape it with `^`.
 for strings and symbols respectively.
 """
 macro df(d, x)
-    esc(_df(d,x))
+    syms = Expr[]
+    vars = Symbol[]
+    plot_call = _df(d, x, syms, vars)
+    compute_vars = Expr(:(=), Expr(:tuple, vars...),
+        Expr(:call, :(StatPlots.compute_all), d, syms...))
+    esc(Expr(:block, compute_vars, plot_call))
 end
 
-_df(d, x) = x
+_df(d, x, syms, vars) = x
 
-function _df(d, x::Expr)
-    (x.head == :quote) && return :(StatPlots.select_column($d, $x))
+function _df(d, x::Expr, syms, vars)
+    if x.head == :quote
+        new_var = gensym(x.args[1])
+        push!(syms, x)
+        push!(vars, new_var)
+        return new_var
+    end
     if x.head == :call
         x.args[1] == :^ && length(x.args) == 2 && return x.args[2]
         x.args[1] == :cols && return :(hcat((StatPlots.select_column($d, i) for i in $(x.args[2]))...))
     end
-    return Expr(x.head, _df.(d, x.args)...)
+    return Expr(x.head, (_df(d, arg, syms, vars) for arg in x.args)...)
 end
+
+compute_all(d, s...) = [StatPlots.select_column(d, ss) for ss in s]
 
 function _argnames(d, x::Expr)
     [_arg2string(d, s) for s in x.args[2:end] if not_kw(s)]
--- a/src/df.jl
+++ b/src/df.jl
@@ -18,7 +18,7 @@ function _df(d, x::Expr)
     (x.head == :quote) && return :(StatPlots.select_column($d, $x))
     if x.head == :call
         x.args[1] == :^ && length(x.args) == 2 && return x.args[2]
-        x.args[1] == :cols && return :(hcat((StatPlots.convert_column($d[i]) for i in $(x.args[2]))...))
+        x.args[1] == :cols && return :(hcat((StatPlots.select_column($d, i) for i in $(x.args[2]))...))
     end
     return Expr(x.head, _df.(d, x.args)...)
 end
@@ -45,18 +45,25 @@ end
 
 stringify(x) = filter(t -> t != ':', string(x))
 
-select_column(df, s) = haskey(df, s) ? convert_column(df[s]) : s
+function select_column(df, s)
+    iterator = getiterator(df)
+    if s in fieldnames(next(iterator, 1)[1]) || isa(s, Integer)
+        return convert_column([getfield(i, s) for i in iterator])
+    else
+        return s
+    end
+end
 
 convert_column(col) = col
 
-function convert_column(col::AbstractDataArray{T}) where T
+function convert_column(col::AbstractArray{T}) where T<:DataValue
     try
-        convert(Array, col)
+        get.(col)
     catch
         error("Missing data of type $T is not supported")
     end
 end
 
-convert_column(col::AbstractDataArray{<:AbstractString}) = convert(Array, col, "")
-convert_column(col::AbstractDataArray{Symbol}) = convert(Array, col, Symbol())
-convert_column(col::AbstractDataArray{<:Real}) = convert(Array, convert(DataArray{Float64}, col), NaN)
+convert_column(col::AbstractArray{DataValue{<:AbstractString}}) = get.(col, "")
+convert_column(col::AbstractArray{DataValue{Symbol}}) = get.(col, Symbol())
+convert_column(col::AbstractDataArray{DataValue{<:Real}}) = get.(convert.(DataValue{Float64}, col), NaN)
--- a/src/df.jl
+++ b/src/df.jl
@@ -49,21 +49,21 @@ function select_column(df, s)
     v = try
             [getfield(i, s) for i in getiterator(df)]
         catch
-            s
+            return s
         end
-    return convert_column(v)
+    return convert_missing.(v)
 end
 
-convert_column(col) = col
+convert_missing(el) = el
 
-function convert_column(col::AbstractArray{T}) where T<:DataValue
+function convert_missing(el::DataValue{T}) where T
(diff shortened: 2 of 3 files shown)


src/StatPlots.jl
```diff
@@ -7,6 +7,9 @@ import Plots: _cycle
 using StatsBase
 using Distributions
 using DataFrames
+using IterableTables
+import DataValues: DataValue
```
mkborregaard: I think you could also implement these from IterableTables - wouldn't that be cleaner?

piever: good point

mkborregaard: I gotta say your macro-fu is pretty on point :-)

mkborregaard:
I think this could be argued to also close https://github.com/JuliaPlots/Plots.jl/issues/53

mkborregaard:
It'd be nice to have some more opinions here, as the table support is really important for StatPlots. The idea is that StatPlots could use IterableTables to 1) support a very general interface to julia's table ecosystem, and 2) drop the bloated DataFrames dep. I like it. Thoughts from other JuliaPlots members? In particular @daschw @tbreloff @oschulz @ChrisRackauckas ?

davidanthoff:
Really excited to see this!

I'll try to look at the code in more detail tomorrow, but I should probably say a word about the relation of [TableTraits.jl](https://github.com/davidanthoff/TableTraits.jl), [TableTraitsUtils.jl](https://github.com/davidanthoff/TableTraitsUtils.jl) and [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl) because that is not yet described in the documentation.

The high level story is that I moved a bunch of things out of [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl), so that packages need to take on less dependencies if they want to integrate with this stack. Here is the new split:
- [TableTraits.jl](https://github.com/davidanthoff/TableTraits.jl) has the very minimal definition of the functions that make up the iterable tables trait. Any package that wants to integrate with iterable tables needs to take a dependency on this one.
- [TableTraitsUtils.jl](https://github.com/davidanthoff/TableTraitsUtils.jl) is a package that provides a default implementation of the iterable table interface. It essentially provides a couple of functions that might make it easier for a given source or sink to integrate with iterable tables. One can use it or not, up to each individual package. That package exports two functions right now: ``create_tableiterator`` you pass it a tuple of arrays and the names of the columns, and it will return an efficient row iterator. This is useful for packages that are sources (so not relevant here). ``create_columns_from_iterabletable`` is for sinks like this here: you pass it an iterable table, and it returns a tuple with the first argument being a tuple of arrays (the column data) and the second element being the names of the columns. [Here](https://github.com/JuliaComputing/IndexedTables.jl/blob/master/src/tabletraits.jl#L34) is an example how it can be used. So any sink that ends up materializing an iterable table into a vector of vectors can use this implementation, which is quite optimized. Not sure whether that would be helpful here or not.
- [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl) has all the integrations with various packages that have not moved into those packages. My hope is that eventually it will go away. For example, we very recently moved the integration for [IndexedTables.jl](https://github.com/JuliaComputing/IndexedTables.jl) out of [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl) and into [IndexedTables.jl](https://github.com/JuliaComputing/IndexedTables.jl). My hope is that we can do that for all the sinks and sources. BUT, my sense is that will not happen anytime soon for [DataFrames.jl](https://github.com/JuliaStats/DataFrames.jl), because there is some mega-refactoring/roadmapping going on over there and my sense is that the maintainers over there don't want to take a dependency on [DataValues.jl](https://github.com/davidanthoff/DataValues.jl) (they are moving things over to [Nulls.jl](https://github.com/JuliaData/Nulls.jl) right now, which is not an option for my stack because it is too slow on julia 0.6 and doesn't work with [Query.jl](https://github.com/davidanthoff/Query.jl) as of right now). So, for right now, someone has to load [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl) so that the [DataFrames.jl](https://github.com/JuliaStats/DataFrames.jl) integration just works... On my end I have packages like [Query.jl](https://github.com/davidanthoff/Query.jl), [CSVFiles.jl](https://github.com/davidanthoff/CSVFiles.jl) etc. load [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl) by default, and I think I'll keep it that way, just to make sure users have a smooth experience... Probably makes sense to do the same here. But the way I would do this is that your complete implementation should only depend on stuff in [TableTraits.jl](https://github.com/davidanthoff/TableTraits.jl) (and maybe [TableTraitsUtils.jl](https://github.com/davidanthoff/TableTraitsUtils.jl)), and then you just put one ``import IterableTables`` somewhere so that the integration with [DataFrames.jl](https://github.com/JuliaStats/DataFrames.jl) is loaded, but don't use anything from [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl) explicitly.

Sorry for the long text, might well be that this was all clear to you guys already :)

davidanthoff:
> I think this could be argued to also close JuliaPlots/Plots.jl#53

Yes, you should be able to consume any DataStreams.jl source via this PR here.

ChrisRackauckas:
I think IterableTables.jl is great and setting StatPlots.jl to work with this ecosystem will only be a good idea in the long run.

src/StatPlots.jl
```diff
@@ -7,6 +7,9 @@ import Plots: _cycle
 using StatsBase
 using Distributions
 using DataFrames
+using IterableTables
```
davidanthoff: I would make this ``import IterableTables``. You really just want the package to be loaded, but you don't need anything from it directly.

src/StatPlots.jl
davidanthoff: Actually, I'd leave it like it is, see my other comments.

src/df.jl
```diff
@@ -45,18 +60,23 @@ end
...
-convert_column(col) = col
+compute_name(df, i) = column_names(IterableTables.getiterator(df))[i]
 
-function convert_column(col::AbstractDataArray{T}) where T
-    try
-        convert(Array, col)
-    catch
-        error("Missing data of type $T is not supported")
+function compute_all(df, syms...)
+    iter = IterableTables.getiterator(df)
```
davidanthoff: You probably want to check whether you actually got an iterable table by calling ``TableTraits.isiterabletable(df)``, right? Someone could have passed you something that is something else entirely.

src/df.jl
```diff
@@ -45,18 +60,23 @@ end
...
+    type_info = Dict(zip(column_names(iter), column_types(iter)))
+    cols = Tuple(isa(s, Integer) ? Array{column_types(iter)[s]}(0) :
+        s in column_names(iter) ? Array{type_info[s]}(0) : s for s in syms)
```
davidanthoff: Ah, so here you actually only materialize some columns, right? I think that means you should not use TableTraitsUtils.jl, because that will always materialize all columns.

One thing you could add later is to check whether ``Base.iteratorsize(typeof(iter))==Base.HasLength()``, in which case you can call ``length(iter)`` to get the number of rows you will eventually get and then you can either pre-allocate the array or maybe just use ``sizehint!`` on the arrays that you allocate here. In general that should speed up the ``for`` loop later where you push into the array.

daschw:
This is great! Nice work @piever! I'm definitely in favour of this change and dropping the DataFrames dependency.
Thanks also @davidanthoff for providing these valuable insights! It was not at all clear to me already 😄

src/df.jl
mkborregaard: Yes, essentially the idea is for the df macro to be a with-style macro that only extracts the columns that are actually needed (and only tries to do this on symbols that are names in the table).

mkborregaard:
Yes, thanks @davidanthoff and once again to @piever for taking the lead on this!

mkborregaard:
With regards to potential conflicts with the Nulls.jl-based architecture I think we can cross that bridge when we come to it.

piever:
Yep, thanks @davidanthoff for the explanation, it was very helpful and at least to me not at all obvious.

src/df.jl
piever: I should probably check all the techniques you use in TableTraitsUtils.jl to speed up this part (it seems to me that mainly I'm missing unlooping the `for` loop over column names with generated functions and preallocating wherever possible). Still, this feels like unnecessary code duplication. By looking at your code, it seems like it'd be very easy to generalize it to only return some of the columns, with an appropriate keyword argument (if you want I can try and make a PR there). In this case, I don't think this would go against the spirit of keeping IterableTables as tiny as possible, as developers of other packages wouldn't need to do anything extra to support this new feature, whereas the StatPlots use case (select a few columns from an IterableTable) seems common enough and I'm not sure the code to do that efficiently belongs to a statistical plotting package. What do you think?

davidanthoff: Yes, I thought a bit more about this and actually came to the same conclusion :) If we just keep that as an option in TableTraitsUtils.jl it is all cool as it won't complicate the core interface in TableTraits.jl at all.

oschulz: @davidanthoff This is the fist time I'm taking a deeper look at IterableTables (has been on my to-do list for a while though ;-) ). Maybe I didn't look in the right place in the docs - what is the recommended way to iterate over a subset of columns without loading all of them?

oschulz: If there isn't anything like that at the moment - @piever,  @davidanthoff, if I understand correctly, you're suggesting something like `getiterator(ds, cols = ...)`? That would be very handy in many other use cases (data analysis in general), indeed.

davidanthoff: I don't want to change the iterable tables interface, i.e. you would still always get an iterator that returns a named tuple that has all the columns in it. But I want to now provide a helper function in TableTraitsUtils.jl that materializes only a subset of the columns into arrays.

There are two reasons: one is that it keeps the iterable tables interface simple. Adding the ability to tell the source which columns it should return adds additional complexity requirements on the source, and it is a bit difficult to see where to stop. Plus, I think I have a vague idea how one can have a similar story based on something in Query.jl. Not fleshed out yet, but stay tuned ;)

The other reason is that I think the compiler might actually be able to optimize all the tuple access to columns that aren't used away. I'm not a 100% sure about that, but when I talked with Keno about that in July I got the impression that he might think that feasible. In general, if one is careful, a lot of things in the iterable tables interface get inlined and one typically ends up with just one tight loop with no function calls at all. If, inside that loop, one accesses a column, but then that column is never used inside the loop, it should in theory be feasible for the compiler to detect that and just remove that access. If that is true and works, there is actually no need to add an option to the interface itself to select columns. I think for now it is probably best to at some point check out that option carefully, and if that can't be pulled of we can revisit an option in the core iterable tables interface.

oschulz: > The other reason is that I think the compiler might actually be able to optimize all the tuple access to columns that aren't used away.

I had hoped that might be the case, but at least in a very simple toy example I tried it wasn't the case. Inlining is, of course, sometimes hard to predict/control.

But I was also thinking about out-of-core data sources (e.g. databases, column-store file-formats, etc.), where one may need up-front knowledge about which columns to load, as loading columns may be expensive. I think in Query.jl, that's currently hard-coded for SQLite, but a generic solution would be very cool.

I do understand the need to keep things simple, of course.

> typically ends up with just one tight loop with no function calls at all

Might that even SIMD-vectorize? Would depend on not having any branch-expressions, of course (so no on-demand loading, etc.).

oschulz: Sorry for getting out-of-topic a bit, here.

davidanthoff: > I had hoped that might be the case, but at least in a very simple toy example I tried it wasn't. Inlining is, of course, sometimes hard to predict/control.

I tried that way, way back and a lot of even quite involved Query.jl queries ended up being inlined  into one tight loop. I'm not sure whether that is enough for the compiler to ignore access to columns that are not needed, though...

> I think in Query.jl, that's currently hard-coded for SQLite, but a generic solution would be very cool.

Yes, totally agree. My current thinking is that it would be great if a source could implement a partial Query.jl backend. For example, say you have this code ``load("file.csv") |> @select({_.a, _.b}) |> DataFrame`` (*), it would first go to a CSVFiles.jl Query.jl backend that realizes that this query implies that only column ``a`` and ``b`` are needed, and then skips all other columns at CSV parse time already. The same logic could apply to lots of other sources. But for this to work it would be really key that CSVFiles.jl doesn't have to implement a full Query.jl backend, but some lighter version of it. I've been mulling this whole question for a while, and it is slowly clearing up in my mind, but will probably still take a while.

> Might that even SIMD-vectorize? Would depend on not having any branch-expressions, of course (so no on-demand loading, etc.).

Yeah, I think that could probably be made to work.

oschulz: Sounds exciting - I look forward to see what you'll come up with. I know how it is, sometimes concepts just need some time to mature in ones mind ...

piever:
I've incorporated @davidanthoff feedback and used the method from https://github.com/davidanthoff/TableTraitsUtils.jl/pull/2 to efficiently materialize a subset of columns from an iterable table. Will merge tomorrow if everybody's ok.

src/df.jl
```diff
@@ -9,18 +9,33 @@ If you want to avoid replacing the symbol, escape it with `^`.
...
+    compute_vars = Expr(:(=), Expr(:tuple, vars...), Expr(:call, :(StatPlots.compute_all), d, syms...))
+    esc(Expr(:block, compute_vars, plot_call))
 end
 
-_df(d, x) = x
+_df(d, x, syms, vars) = x
 
-function _df(d, x::Expr)
-    (x.head == :quote) && return :(StatPlots.select_column($d, $x))
+function _df(d, x::Expr, syms, vars)
```
mkborregaard: Shouldn't this be `_df!` seeing as it modifies the input?

piever: Good point!

mkborregaard: Could the names also be more informative? compute_all seems very broad (should it perhaps be extract_columns_from_table?) and _df a little sparse (parse_table_call or something like that)?

piever: Good choice of names, I'll just replace `table` to `iterabletable` to be consistent with the TraitsUtils functions.

mkborregaard:
:tada: time for a new StatPlots release? Or do we want to play around with it a little first?

piever:
I'd be in favor of releasing quickly, as I was planning to start making potentially breaking changes to `groupapply` and I think it's better to have a release first.

mkborregaard:
I've made the release.

PR #83, opened 2017-08-31
Allow custom grouping behavior for groupedbar

piever:


mkborregaard:
@piever I'm getting the feeling that with these new changes groupapply should be much more simple and intuitive to implement - do you not think so too? The reason I suggested plotting with a two-stage process earlier was that the interface seemed slightly off how other things in Plots worked. Do you think it could be integrated more now so it could be a one-stage userplot? (and more importantly, do you think that would be desirable, given your experience with it? I still really like the idea that you should produce objects during data analysis, and then plot those objects, rather than hiding the analysis element behind an automagical screen - but I guess these two paradigms need not conflict). Thoughts?

piever:
Yes, that's one of the reasons why I'm working so hard on these changes: with proper ribbon plots and grouping in base Plots and `groupedbar` and the `@df` macro in StatPlots, it really simplifies a lot. I would also like to keep the option to save the statistical object, especially because if you have a lot of data, you don't really want to run all the analysis again just to change line width. But the verbosity of the command can probably be cut down a lot (I think the macro can help) and the statistical object itself can also be simplified. I'll open a separate issue to discuss this once everything is merged.

PR #91, opened 2017-09-09
added docs for general table support and group in groupedbar

piever:


README.md
```diff
@@ -171,6 +173,11 @@ groupedbar(rand(10,3), bar_position = :dodge, bar_width=0.7)
 
 ![tmp](https://cloud.githubusercontent.com/assets/933338/18962092/673f6c78-863d-11e6-9ee9-8ca104e5d2a3.png)
 
+The `group` syntax is also possible in combination with `groupedbar`:
+
+```julia
+groupedbar(rand(6), group = [1, 1, 2, 2, 3, 3])
```
mkborregaard: Sorry for commenting after merging, but I think any examples with groupedbar need the x values too. Without x values, the bar group that each bar belongs to will be determined by the order of the data in the original data set, which rarely makes sense.

piever: Agreed, and actually I hadn't noticed that with the latest Query it is possible to combine queries and plots in a really nice way, I think I should add that as well.

PR #96, opened 2017-09-19
more docs

piever:
Added a better `groupedbar` example (see [this discussion](https://github.com/JuliaPlots/StatPlots.jl/pull/91#discussion_r138714961) ) and mentioned the possibility of combining plot commands with Query pipelines. In the future we may also think whether to change the `@df` macro so that it combines even more smoothly with queries (for example, should it have a curryied form? should we also add Query syntax for selecting columns inside `@df` and so on).

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,12 +44,11 @@ df[:red] = rand(10)
 The `@df` macro plays nicely with the new syntax of the [Query.jl](https://github.com/davidanthoff/Query.jl) data manipulation package (v0.7 and above), in that a plot command can be added at the end of a query pipeline, without having to explicitly collect the outcome of the query first:
 
 ```julia
-using Lazy, Query, StatPlots
-@> df begin
-    @where _.a > 5
-    @select {_.b, d = _.c-10}
+using Query, StatPlots
+df |>
+    @where(_.a > 5) |>
+    @select({_.b, d = _.c-10}) |>
     @df scatter(:b, :d)
-end
 ```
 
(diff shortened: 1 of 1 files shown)


piever:
Basically what happens is that function concatenation with `|>` which is available in Base only works with functions of one argument, which `@df` is not, as it takes 2 argument to make sense. To solve this, there is the concatenation macro `@>` which concatenates passing the outcome of a function call as the first argument of the following function call, so that

```julia
@> df begin
    @where _.a > 5
    @select {_.b, d = _.c-10}
    @df scatter(:b, :d)
end
```

is actually equivalent to:

```julia
@df(@select(@where(df, _.a > 5), {_.b, d = _.c-10}), scatter(:b, :d))
```

but much more readable. Query.jl used a different approach which works with standard concatenation, that is to say it provides explicit Curryed methods, by definining:

```julia
where(y) = x -> where(x, y)
```

Personally I'm not a fan of this approach, and I find the `@>` macro more elegant than a bunch of `|>` operators but I agree that if we want to be as consistent as possible with Query, we should also add a Curryed method for `@df` and change the example to use this type of syntax:

```julia
using Query, DataFrames

df = DataFrame(a=[1,1,2,3], b=[4,5,6,8])

df2 = df |>
    @groupby(_.a) |>
    @select({a=_.key, b=mean(_..b)}) |>
    @where(_.b > 5) |>
    @orderby_descending(_.b) |>
    DataFrame
```

mkborregaard:
I think that would be clearer in this case.

piever:
Ok, I'll add the Curryed version of `@df` in a separate PR and then update the syntax here then.

PR #97, opened 2017-09-19
fixes #94 correctly catch datavalue error

### LINKED ISSUE #94: "Missing data of type Any is not supported" for df macro.
Consider the following piece of code:
```julia
using StatPlots, DataFrames

df_any = DataFrame(Any, 0, 1)
df_array_any = DataFrame(Array{Any}(0, 1))

for i in 1:10
    push!(df_any, rand())
    push!(df_array_any, rand())
end

@df df_any plot(:x1) # Error: "Missing data of type Any is not supported"
@df df_array_any plot(:x1) # works
```
I'm not sure if this is related to StatPlots or DataFrames, if this is expected or an issue. I was just a little surprised.
cc: @piever

piever:
Now

```julia
using StatPlots, DataFrames

df_any = DataFrame(Any, 0, 1)

for i in 1:10
    push!(df_any, rand())
end

@df df_any plot(:x1)
```

works, as it should, and it gives the correct error message if there is missing data we don't know how to handle:

```julia
df_any[2, :x1] = NA
@df df_any plot(:x1)    # Error: "Missing data of type Any is not supported"
```

### DIFF
--- a/src/df.jl
+++ b/src/df.jl
@@ -74,7 +74,7 @@ function extract_columns_from_iterabletable(df, syms...)
 end
 
 convert_missing(el) = el
-convert_missing(el::DataValue{T}) where {T} = get(el, error("Missing data of type $T is not supported"))
+convert_missing(el::DataValue{T}) where {T} = isnull(el) ? error("Missing data of type $T is not supported") : el.value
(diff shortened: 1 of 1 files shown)


daschw:
Great!

PR #99, opened 2017-09-28
RFC: add qqnorm recipe

mkborregaard:
Adds a qqnorm and qqplot recipe

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,11 +176,12 @@ x = rand(Normal(), 100)
 y = rand(Cauchy(), 100)
 
 plot(
- qqplot(x, y),
- qqnorm(x, smooth = true)
+ qqplot(x, y, qqline = :fit), # qqplot of two samples, show a fitted regression line
+ qqplot(Cauchy, y),           # compare with a Cauchy distribution fitted to y; pass an instance (e.g. Normal(0,1)) to compare with a specific distribution
+ qqnorm(x, qqline = :R)       # the :R default line passes through the 1st and 3rd quartiles of the distribution
 )
 ```
-<img width="788" alt="skaermbillede 2017-09-28 kl 08 36 56" src="https://user-images.githubusercontent.com/8429802/30963544-1fa789ec-a44e-11e7-9d40-2ef4f33463e6.png">
+<img width="1185" alt="skaermbillede 2017-09-28 kl 22 46 28" src="https://user-images.githubusercontent.com/8429802/30989741-0c4f9dac-a49f-11e7-98ff-028192a8d5b1.png">
 
 ## Grouped Bar plots
 
--- a/README.md
+++ b/README.md
@@ -176,11 +176,11 @@ x = rand(Normal(), 100)
 y = rand(Cauchy(), 100)
 
 plot(
- qqnorm(x),
- qqplot(x, y)
+ qqplot(x, y),
+ qqnorm(x, smooth = true)
 )
 ```
-<img width="788" alt="skaermbillede 2017-09-28 kl 08 36 56" src="https://user-images.githubusercontent.com/8429802/30952479-4f4dc18c-a428-11e7-9856-0601cb420fba.png">
+<img width="788" alt="skaermbillede 2017-09-28 kl 08 36 56" src="https://user-images.githubusercontent.com/8429802/30963544-1fa789ec-a44e-11e7-9d40-2ef4f33463e6.png">
 
 ## Grouped Bar plots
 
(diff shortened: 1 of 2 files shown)


mkborregaard:
RFC: Should we show a fitted line behind the points? (Or a 0,1 line?). Should the `qnorm` plot against N(0,1) or against N(μ, σ) where μ and σ are estimated from the empirical distribution?

daschw:
> Should the qnorm plot against N(0,1) or against N(μ, σ) where μ and σ are estimated from the empirical distribution?

How about defaulting to N(0,1) and allowing μ and σ to be passed as optional kwargs in the recipe? Or another flag that says: please estimate μ and σ for me?

mkborregaard:
Could be - but `qqnorm(x)` is just a default shorthand for `qqplot(Normal(μ, σ), x)` or qqplot(Normal(0, 1), x)` in the first place, which seems just as easy to write if you want full control.

daschw:
OK, I see.
Maybe I get this wrong, but if you decide to default to N(0,1), but the users want to have μ and σ estimated, they cannot write `qqplot(Normal(μ, σ), x)`, because they do not know μ and σ. On the other hand, if the default is estimating μ and σ, they can instead easily write `qqplot(Normal(0, 1), x)`.

mkborregaard:
good point for sure!

piever:
Very nice!
In terms of adding the line, I think it should be an optional keyword argument, and one should be able to specify if one wants identity line of fitted line. Both putting the identity line (if the questions is "are X and Y from the same distribution?") and putting a fitted line (if the question is "is Y a rescaled/shifted" version of X) make sense. I think R has a specific `qqline`function but I think a keyword should do.

I also think we should support theoretical distributions (which we do already in this PR I guess?), so that `qqnorm(X) = qqplot(Normal(), X)`. We could also support distribution families, for example one way to address @daschw comment would be:

```julia
qqplot(D::Type{T}, X) where {T<:Distribution} = qqplot(fit(D, X), X)
```

so that `qqplot(Normal, X) == qqplot(Normal(μ, σ), X)`, (where `μ, σ` are the fitted parameters) but it'd be equally easy to do that for any other distribution family.

As to what should `qqnorm` default to, I think R does `qqplot(Normal(0, 1), X)` and we should probably just be in line with what other people do.

mkborregaard:
We already have a keyword for the fitted line, `smooth = true`.

piever:
It's true, you could use `smooth = true` :). I agree that the behavior would be very different, but if it's common enough to fit a distribution and then check the quality of the fit with a qqplot, then it'd be nice to have this. It's also true that writing `qqplot(fit(Normal, X), X)` is not too bad either.

mkborregaard:
Muarhaha, so I did some reading and it appears that
1. R uses N(0,1) (and the Wikipedia article seems to recommend that)
2. Stata, SPSS and Matlab uses N(μ, σ)

with regards to the line
1. Matlab uses a trend line
2. Python appears to use the 0,1 line
3. R draws a line between a predefined set of quantiles, defaulting to (0.25 and 0.75)

So - I guess you can say it's not surprising we have some different ideas...

Here is my suggestion:
To use N(μ, σ) for `qqnorm` because it easy to `qqplot(Normal(), x)` as @daschw stated, and to plot the 0,1 line (because that is the best test when the distribution is fitted). There should be a qqline keyword that could take :identity (the default), :fit (just switch smooth to true), and :quantile, ignoring the possibility of setting alternative quantiles.

I'll add @piever's suggestion of fitting when you supply the type.

?

daschw:
Sounds great!

piever:
Ups, I proposed to follow what other languages do and only checked R which is an outlier...
The plan sounds perfect, but you are right that it's important to stress in the docs that:
`qqplot(Normal, X) != qqplot(Normal(), X)`

mkborregaard:
Fixed. The only small hiccup being that `qqline = :fit` gives a trendline of the same color as the data points, whereas the other options are a new series and give a line of a different color. I could make the other options give the same color as well by passing to a series recipe, but I kind of like the different color here.

mkborregaard:
Right - ready to merge now. Newest readme example:

```julia
x = rand(Normal(), 100)
y = rand(Cauchy(), 100)

plot(
 qqplot(x, y, qqline = :fit), # qqplot of two samples, show a fitted regression line
 qqplot(Cauchy, y),           # compare with a Cauchy distribution fitted to y; pass an instance (e.g. Normal(0,1)) to compare with a specific distribution
 qqnorm(x, qqline = :R)       # the :R default line passes through the 1st and 3rd quartiles of the distribution
)
```
skaermbillede 2017-09-28 kl 22 46 28

mkborregaard:
Thanks for the comments, merging now :-)

PR #100, opened 2017-10-03
Deprecate groupapply and introduce the GroupedErrors package

piever:


piever:
Cool! I'm merging with the added "highly recommended". Should we say [here](https://github.com/JuliaData/DataFrames.jl/issues/1232) that we don't need any update for the new DataFrames? With `@df` and GroupedErrors we don't rely on any specific nullability implementation

PR #107, opened 2017-10-25
Update README.md - add groupedbar example

kaslusimoes:
Just adding the example @mkborregaard built from the discussion on [discourse](https://discourse.julialang.org/t/plots-bars-side-by-side/6655/4)

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,20 @@ The `group` syntax is also possible in combination with `groupedbar`:
 ```julia
 groupedbar([1, 2, 1, 2, 1, 2], rand(6), group = [1, 1, 2, 2, 3, 3])
 ```
+A more complex example is also possible:
+
+```julia
+mn = [20, 35, 30, 35, 27,25, 32, 34, 20, 25]
+sx = repeat(["Men", "Women"], inner = 5)
+std = [2, 3, 4, 1, 2, 3, 5, 2, 3, 3]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -204,6 +204,20 @@ The `group` syntax is also possible in combination with `groupedbar`:
 ```julia
 groupedbar([1, 2, 1, 2, 1, 2], rand(6), group = [1, 1, 2, 2, 3, 3])
 ```
+A more complex example is also possible:
```
mkborregaard: Can you delete this and the example before? And consider whether one of the two previous graphs could be taken out as well? I don't think we need more than two plots for groupedbar, and they should ideally be combined to a two-panel plot (perhaps reusing the data, but letting e.g. one represent the column-matrix attributes and the minimum input, the other the grouped method).

mkborregaard:
Thanks. It'd be great, once we're editing it, to rethink all of the groupedbar example rather than just stacking new info on top.

README.md
kaslusimoes: Sorry, I cannot understand the example you are giving. Can you please elaborate further?

Either way I already deleted this line and the example before.

mkborregaard:
Totally forgot about this, sorry. Thanks!

kaslusimoes:
Hehe that's no problem. Thank you for merging!

PR #110, opened 2017-11-17
suppress axes labels in marginalhists

daschw:


piever:
Nice! This closes #79

PR #111, opened 2017-11-27
Deprecate old DataFrames syntax

mkborregaard:
This PR deprecates the `plot(df, :x, :y)` syntax, so that the `@df plot(:x, :y)` is the only valid syntax for DataFrames.
It also removes shadederror and groupederror which have both been moved to the GroupedErrors package


mkborregaard:
@piever to be compliant with the new syntax, all we'd need to do is `import Missings` and replace `isnull` with `ismissing` here https://github.com/JuliaPlots/StatPlots.jl/blob/master/src/df.jl#L87 or?

src/StatPlots.jl
```diff
@@ -34,8 +32,6 @@ include("violin.jl")
 include("hist.jl")
 include("marginalhist.jl")
 include("bar.jl")
-include("shadederror.jl")
```
mkborregaard: Let me know if this is too heavy-handed, but the syntax has moved out, no?

piever: I agree with removing it. It didn't really move anywhere but became superfluous after all the PRs to fix ribbon plots, so now the recommended way is `plot(args...; ribbon = ...)` which is much more standard.

piever:
To adjust to the new syntax I don't think we need to do anything... The `isnull` that we call is the `isnull` from DataValues. The Query/ IterableTables universe uses DataValues as missing data, unlike DataFrames which uses Missings: it is extremely confusing but there are some technical obstacles that don't allow unification. I'll ask on Slack if it's still going to work, but I think it should, in which case we're good as it is.

mkborregaard:
Awesome, thanks a lot. If you have no concerns I'll merge and push a release.

piever:
Yes, looks good to me.

PR #115, opened 2017-12-01
enable tick labels in last row of cornerplot (fix #114)

### LINKED ISSUE #114: Cornerplot shows no x-ticks along the bottom axis
figure_1-1
There are no x-tick labels along the bottom plots.  However, they should be there, as they are enabled in this [line](https://github.com/JuliaPlots/StatPlots.jl/blob/7cccc75f450cecb86bef723ab4ab17a5964163a6/src/cornerplot.jl#L93) (as far as I understand).

daschw:
I'm still not entirely sure, why the original implementation did not work (cf #114). This however fixes the problem:
cornerplot

### DIFF
--- a/src/cornerplot.jl
+++ b/src/cornerplot.jl
@@ -44,7 +44,8 @@
     linecolor --> Plots.fg_color(plotattributes)
     grid      --> true
     ticks     := nothing
-    formatter := v->""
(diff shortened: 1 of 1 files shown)


mauro3:
Thanks! Also note, the original used to work but then stopped sometime after summer.

PR #117, opened 2017-12-03
added labelling to df

piever:
Now the `@df` macro also fills in the legend when there are multiple series:

```julia
df = DataFrame(a = 1:10, b = 10*rand(10), c = 10 * rand(10))
@df df plot(:a, [sqrt.(:b) :c])
```
label

```julia
@df iris corrplot(cols(1:4))
```

corrplot

I didn't add automatic labelling of axis so far because:

- We don't normally write anything on the axis (whereas we add `y1, y2, ...` on the legend)
- It's hard to infer what is on what axis when there only is one variable from the function call. For example `@df df scatter(:a)` should label the `y` axis whereas `@df df density(:a)` should label the `x` axis

I'm wondering whether some of it could also be added in Plots with a `@label` macro that would add the legend to a plot (see [#1072](https://github.com/JuliaPlots/Plots.jl/issues/1072)).

### DIFF
--- a/src/df.jl
+++ b/src/df.jl
@@ -42,13 +42,14 @@ end
 parse_iterabletable_call!(d, x, syms, vars) = x
 
 function parse_iterabletable_call!(d, x::Expr, syms, vars)
-    if x.head == :quote
+    if x.head == :. && length(x.args) == 2
+        isa(x.args[2], Expr) && (x.args[2].head == :quote) && return x
+    elseif x.head == :quote
         new_var = gensym(x.args[1])
         push!(syms, x)
         push!(vars, new_var)
         return new_var
-    end
-    if x.head == :call
+    elseif x.head == :call
         x.args[1] == :^ && length(x.args) == 2 && return x.args[2]
         if x.args[1] == :cols
             range = eval(x.args[2])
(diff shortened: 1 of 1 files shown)


mkborregaard:
With regard to your first point above, I do think it'd be nice to fill in the axes automatically here. The more I think about it, the more I think your previous idea of having `@df` insert a reserved keyword for the axes which is handled in Plots when the final seriestype is known is the only approach to your second point.

piever:
I'll probably merge this as a stop gap solution and try to figure out how to add the new keyword in Plots in the longer term for axis labelling (should only be required for 1 arg plots).

piever:
There's one thing on which I could use some feedback before merging. I've seen on PlotThemes the syntax:

```julia
@df singers begin
    violin(:VoicePart, :Height)
    boxplot!(:VoicePart, :Height)
end
```

This PR would break that syntax as it assumes that the second argument of the `df` macro is a function call (to add to it the extra keyword argument `label`). Do we think we want to keep the "block syntax" as a possibility? In that case, the best I could do (to also add the name) would be to define the macro to apply itself on each line of the block if the second argument is a block. For example:

```julia
@df singers begin
    violin(:VoicePart, :Height)
    boxplot!(:VoicePart, :Height)
end
```

would return:

```julia
@df singers violin(:VoicePart, :Height)
@df singers boxplot!(:VoicePart, :Height)
end
```

mkborregaard:
I think that sounds pretty ideal?

PR #118, opened 2017-12-06
support discrete distributions; fix #116

### LINKED ISSUE #116: Non-Real valued distributions
The Recipe for Distributions does not work with distributions with `Integer` domains.
Like `Distributions.Bernoulli`, `Distributions.BetaBinomial `, etc..

mkborregaard:


mkborregaard:
It sort of works, but setting :steppost as the default seriestypes is not without issues, as can be seen here:

```julia
using StatPlots, Distributions
n = Normal(4,5)
d = BetaBinomial(10, 0.3, 2)
plot([n, d])
```
skaermbillede 2017-12-06 kl 11 08 04

mkborregaard:
Maybe having bars could be considered:
skaermbillede 2017-12-06 kl 11 11 18

AShedko:
Looks good.

mkborregaard:
@daschw  @piever any opinions on the preferred seriestype for discrete distributions, for best looks/consistency ?
```julia
N = Normal(50, 10)
P = Poisson(50)
plot(
       plot(N),
       plot(N, fill = 0),
       plot(P, st = :bar),
       plot(P, st = :bar, lw = 0),
       plot(P, st = :steppost),
       plot(P, st = :steppost, fill = 0), 
       legend = false, size = (800, 1000), layout = (3,2))
```
skaermbillede 2017-12-06 kl 13 16 04

piever:
I'm definitely against `:steppost` as the distribution only has a non-zero density at the point of integer coordinates, instead `steppost` would suggest that it has a pdf with the same value in say 4 and 4.5. I would instead consider a scatterplot. See for example the [Poisson distribution](https://en.wikipedia.org/wiki/Poisson_distribution) from the wikipedia page.

mkborregaard:
So do you suggest scatter + line? And is that better than :bar?

piever:
I'd go for scatter + line. Bar is a perfectly viable alternative but how would it look like if I plotted two Poisson distributions on the same plot? Would one just cover the other by default?

mkborregaard:
I've added a `:scatterpath` recipe (works better than fiddling with markershape IMHO), now it looks like
```
p = Poisson(15)
plot(p)
```
skaermbillede 2017-12-06 kl 23 16 40

PR #150, opened 2018-05-29
Deal with NaN's in y

lyon-fnal:
This PR fixes issue #149

### DIFF
--- a/src/bar.jl
+++ b/src/bar.jl
@@ -48,7 +48,7 @@ grouped_xy(y::AbstractMatrix) = 1:size(y,1), y
     fillrange := if isstack
         # shift y/fillrange up
         y = copy(y)
-        y[isnan.(y)] = 0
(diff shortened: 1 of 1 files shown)


src/bar.jl
```diff
@@ -48,6 +48,7 @@ grouped_xy(y::AbstractMatrix) = 1:size(y,1), y
     fillrange := if isstack
         # shift y/fillrange up
         y = copy(y)
+        y[isnan.(y)] = 0
```
mkborregaard: Perhaps better `!isfinite.(y)`?

lyon-fnal: That works, though the deprecation warning said to do `y[.!isfinite.(y)] = 0`. I'll update the PR.

mkborregaard:
Thanks a lot @lyon-fnal !

lyon-fnal:
Happy to help!

PR #152, opened 2018-07-27
hclust dendrogram based on kescobo's implementation in Microbiome.jl

funnell:
Based on implementation by @kescobo with some refactoring. @kescobo, if you'd prefer to make your own pull request, please let me know! I just wanted something in StatPlots for some other work I'd like to do, but don't want to steal credit.
Otherwise, feedback is appreciated; I've left the useheight parameter in although I'm not sure what the use case is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,18 @@ groupedbar(nam, rand(5, 2), group = ctg, xlabel = "Groups", ylabel = "Scores",
 
 ![](https://user-images.githubusercontent.com/6645258/32116755-b7018f02-bb2a-11e7-82c7-ca471ecaeecf.png)
 
+## Dendrograms
+
+```julia
+using Clustering
+D = rand(10, 10)
+D += D'
+hc = hclust(D, :single)
+plot(hc)
+```
+
+![dendrogram](https://user-images.githubusercontent.com/381464/43355211-855d5aa2-920d-11e8-82d7-2bf1a7aeccb5.png)
+
 ## GroupedErrors.jl for population analysis
 
 Population analysis on a table-like data structures can be done using the highly recommended [GroupedErrors](https://github.com/piever/GroupedErrors.jl) package.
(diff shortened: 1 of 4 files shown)


piever:
Thanks! Even though I imagine that usage is as simple as `plot(h)` where `h` is your `HClust` object, it'd be nice to have one example in the README (a few lines of code + a png of a plot), pretty much like we do for [distributions](https://github.com/JuliaPlots/StatPlots.jl#distributions).

Taking the dependency on Clustering is fine, I think the consensus is that the [list](https://github.com/JuliaStats/Stats.jl/pull/4/files) of packages in the future Stats.jl are all acceptable dependencies.

kescobo:
@funnell No sweat, thanks for taking the time! Happy to have this move out of my package for sure, and have no compunction about you getting the credit for the move :-P.

Agree that example docs would be useful.  I've got some example plots at the [bottom here](https://biojulia.net/Microbiome.jl/stable/distances.html), though I think you can probably make things more visually impressive (that's more to demonstrate the optimal leaf ordering). 

The idea of the `useheight` param is because branch length is important in biological settings. Are you wondering when you'd want it to be `false`?

funnell:
How does this look?
```julia
using Clustering
D = rand(10, 10)
D += D'
hc = hclust(D, :single)
plot(hc)
```
hclust_dendrogram

kescobo:
Agreed! I'd be happy to add a heatmap example to this or in a separate PR

piever:
Sure, the heatmap can also be added later. Let me know when you are both happy with the PR so that I can merge it.

funnell:
My motivation for this PR was to allow the addition of a clustered heatmap plot to StatPlots. My hope is to eventually have a heatmap plot that can handle row and column annotations similar to pheatmap or ComplexHeatmap in R. @kescobo a new PR with a clustered heatmap would be great!

funnell:
I'm ok with this being merged now, unless there are any further suggestions.

piever:
Sure, I'll merge tomorrow if nobody objects.

PR #153, opened 2018-07-28
Add interactive GUI

piever:
This is a work in progress (though I'd like to merge soonish marked as "experimental" to get feedback).

It creates a small GUI starting from a table to create plots interactively. The user can select one or more columns as first variable, then one or more columns as second variable (or none), plot type and whether to split by some columns:

```julia
import RDatasets
iris = RDatasets.dataset("datasets", "iris")
using StatPlots, Interact
interactstats(iris)
```

interactstats

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,22 @@ The old syntax, passing the `DataFrame` as the first argument to the `plot` call
 
 ---
 
+## Visualizing a table interactively
+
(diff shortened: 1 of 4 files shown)


piever:
I'm also unsure about the name, `interactstats` is a bit tricky to spell, would be happy if anybody has better ideas.

PR #160, opened 2018-08-11
WIP: extend recipetype

piever:


piever:
EDIT: maybe better to do this inside the `@userplot` macro directly

mkborregaard:
@piever why did you close this? What was better to do instead? I'm needing the functionality :-)

piever:
Ah, now I remember: I thought the `@userplot` macro could also automatically add this definition, OTOH that''s probably a bad idea (because of name clashes it might be best to keep this "opt-in") so this PR may well be the way to go.

mkborregaard:
Yeah makes sense - sorry for being slow but what are yuo thinking about in the context of name clashes?

mkborregaard:
I'll merge this for now - should be easy to update if we decide to change the userplot behaviour in RecipesBase

Issue #138, opened 2018-04-08
Remove dependency on Plots.jl


juliohm:
What is the part of the code that currently depends on Plots.jl? Could it be purged out or refactored?

I opened an issue a while ago arguing that this package could be split into many smaller packages, and I still think that this argument holds today :+1:

mkborregaard:
I disagree. On the contrary, I think this whole package could be merged into Plots.jl proper. The original purpose of splitting it out was to avoid the heavy dependency of DataFrames, but it no longer has that, and `Missing` support is moving to Base in 0.7.
For the record, it was PlotRecipes, not StatPlots that you suggested splitting. I still think, as stated here https://github.com/JuliaPlots/PlotRecipes.jl/issues/32#issuecomment-322065222 that it would make sense to split the Graph recipes off in their own package, but keep the rest together.

juliohm:
Thank you for sharing. I will rewrite some of the recipes myself on my packages, cannot afford such big dependency at the moment.

piever:
Could you check if you can circumvent this by using Requires and returning for example the`GroupedBar` type in your recipes?

piever:
What I proposed was to have a small package containing only the type definitions for type recipes, which would be a developer target, Plots would then implement the actual recipe. In this way you don't increase your package load time but your recipes would work when the user also loads Plots, which they have to anyway.

juliohm:
Sorry, I misunderstood it :) You mean a package just exporting the names of the recipes without the actual implementation?

piever:
No problem :) Yes, that's the idea. I think currently mainly type recipes are problematic and as they are implemented by overloading some functions, the overloading can happen in a separate package who would have all the necessary dependencies but the type definition would be in the small dependency free package (it's type piracy but I don't think that would be an issue in this case as long as the rules are clear).

There is an other possible strategy. AFAIU to use `groupedbar` in your recipes you basically need to return a `GroupedBar` object. Could you simply use Requires.jl and wrap your recipe in a `@require StatPlots`block and then in that block you can return `StatPlots.GroupedBar` objects? In this way that code would only be loaded when the user does:

```julia
using StatPlots, YourPackage
```

If this works, maybe that's also an acceptable solution.

I'll leave the issue open as making type recipes available without depending on the package is an important unresolved issue.

juliohm:
[...]

PR #161, opened 2018-08-11
Explicitly state that module should not be precompiled

mkborregaard:
Precompilation breaks StatPlots, possibly due to Reexport

### DIFF
--- a/src/StatPlots.jl
+++ b/src/StatPlots.jl
@@ -1,4 +1,4 @@
-
+__precompile__(false)
 module StatPlots
 
 using Reexport


cstjean:
It's a bummer for downstream packages. Is `@reexport Plots` even a good idea? I suppose the point is to save six keystrokes for interactive use?

piever:
It was done to get StatPlots to work on 1.0 as quickly as possible, now that things have calmed down a little bit we can see whether it's still necessary or if there are alternatives.

mkborregaard:
I've come to think that StatPlots (and PlotRecipes) should not depend on Plots at all

piever:
This was mentioned already in #138. It may well be the best solution.

PR #163, opened 2018-08-11
fix julia 1

piever:


mkborregaard:
great - is the foldl deprecation not from StatPlots?

mkborregaard:
oh it's from Widgets

piever:
Widgets is fixed but I need WebIO to upgrade before I can tag

PR #173, opened 2018-08-20
wip: reenable gui

piever:


REQUIRE
```diff
@@ -9,7 +9,7 @@ IterableTables 0.5
 TableTraitsUtils 0.1
 TableTraits
 DataValues
-#Widgets
-#Observables
+Widgets 0.4.0
```
mkborregaard: I think in principle if you don't use any new API developments in Widgets and Observables but simply need them to be 0.7 compliant you won't need to lower-bound them.

piever: In principle yes, but actually need some new `struct` that I've only defined recently (`AbstractObservable`) so I think it's safer to lower bound.

piever:
Merging as I think this is correct and works in Blink but I'll keep the caveat in the README as this may require some notebook fixes to work in the notebook.

PR #182, opened 2018-09-07
Multiple columns in groupedbar

sebastianskejoe:
This adds the ability to do
```
@df d groupedbar(:y, [:x1, :x2, :x3], bar_position=:dodge)
```
instead of manually having to manually transform the desired columns to a matrix.


mkborregaard:
Do you mean instead of having to go `[:x1 :x2 :x3]`? I don't really see the improvement.

sebastianskejoe:
I completely agree. Being new to Julia, I assummed [x1 x2] was the same as [x1, x2]. Thanks for the quick reply, I will close the pull request.

PR #183, opened 2018-09-10
Convert isnull to 1.0 equivalent

opahk:
Convert `isnull` call to the equivalent ` === nothing` comparison in Julia 1.0

### DIFF
--- a/src/df.jl
+++ b/src/df.jl
@@ -165,7 +165,7 @@ function extract_columns_from_iterabletable(df, syms...)
 end
 
 convert_missing(el) = el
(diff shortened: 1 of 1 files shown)


mkborregaard:
Thank you!

piever:
I don't understand, `isnull` is from the DataValues packages and checks if data is missing or not, what does `nothing` have to do with this?

mkborregaard:
oh shoot, I just merged without thinking, as I don't know anything about DataValues.

mkborregaard:
I've reverted it.

Issue #145, opened 2018-04-27
Violin / boxplots should behave like groupedbar w/r/t groups


kescobo:
Not sure how difficult this would be, but violin and boxplot groups would make far more sense if the default were analogous to `bar_position=:dodge` for grouped bars (it's the default for `groupedbar()` atm). 

```
using StatPlots

df = DataFrame(x=repmat([1,2,3],4), y=rand(12), g=vcat(fill("a", 6), fill("b", 6)))
@df df groupedbar(:x, :y, group=:g)
```

![bar](https://user-images.githubusercontent.com/3502975/39387824-dc2280f8-4a49-11e8-8352-962b1bb317e6.png)

```
@df df violin(:x, :y, group=:g)
```

![violin](https://user-images.githubusercontent.com/3502975/39387830-e4f0bc9a-4a49-11e8-8136-c164d2df204d.png)

```
@df df boxplot(:x, :y, group=:g)
```

![box](https://user-images.githubusercontent.com/3502975/39387836-ef9d87ea-4a49-11e8-8b47-d320e3e06766.png)

I looked at the code for the `groupedbar()` user recipe, and it seems fairly straightforward, but the code for `violin()` and `boxplot()` was too hard for me to grok at 6pm on a friday. My main hesitation about wading into them is that it looks like they're not user recipes the way that `groupedbar()` is, and I don't know the implications of converting them.

piever:
If you make a PR I'd be happy to review it, this thing has been on the to do list for a while. It's not completely straightforward, what I would recommend is to start making some `groupedboxplot` and `groupedviolin` that behave as you want as user recipes (just like `groupedbar`). I believe the best way to give them data is probably as two tuples of vectors (one tuple with the x values for each group and another with the y values for each group): the matrix approach here makes less sense as different columns may have different lengths. You could try copying the `groupedbar` approach as much as possible but each series should be `:boxplot` rather than `:bar`.

Then we'll see how to make it so that `groupedboxplot` with a grouping argument calls `groupedboxplot` with the tuples (we should do a similar trick as with `groupedbar`, but shouldn't be too difficult).

kescobo:
I agree with @mkborregaard - the main barrier is the introduction of a new keyword right? In the case of grouped bar, it's the `bar_position` keyword, though I'm not actually sure that would be necessary for violin and boxplots (I've never seen the `stacked` form).

mkborregaard:
[...]

Issue #89, opened 2017-09-07
Violine plot with not overlapping groups


mkrenn:
Hi!
Is there a way to create grouped violin plots on a categorical x-axis, where the violins of different groups on one category do not overlap?
Similar behavior has been discussed in #56 for boxplots, but i didn't find a solution for that.

`df = DataFrame(accuracy=Float64[],Cat1=String[],Cat2=String[])`
`for i = 1 : 100
        push!(df,[rand() rand(["A"; "B"]) rand(["1" "2" "3"])])
end`
 `violin(df,:Cat1,:accuracy,group=:Cat2,ylims=(0,1))`

![violinexample](https://user-images.githubusercontent.com/12136570/30154425-2048bb5e-93ba-11e7-964a-3bf12d66435f.png)

mkborregaard:
Not at the moment. This is similar to `groupedbar` just for violins, and has also been requested for boxplot (https://github.com/JuliaPlots/StatPlots.jl/issues/56). It should be able to generalize that solution.

piever:
This kind of request is being asked more and more often. Maybe it'd be good to factor all the `groupedbar` code that's needed to plot things side by side and then use it to create `groupedviolin`and `groupedboxplot`.

mkborregaard:
👍

mkborregaard:
Superseded by #145

Issue #56, opened 2017-06-07
How to produce a boxplot with groups, where the groups are side-by-side


bisraelsen:
I am making a box plot grouped by a certain variable. By default the boxes appear overlapping on each other. Is there a way to have them side-by-side to avoid the overlap?

mkborregaard:
This is much easier to adress if you post an MWE and a plot. 
The short answer, though, is that this cannot be done automatically at the moment. But that it'd be nice to have. If you do post an MWE I might suggest a workaround.

kescobo:
@mkborregaard is problem related to #44? I'm suspecting I will need this functionality more in the nearish future. I managed to figure out a work-around, and I'd be willing to make an attempt to formalize it into a PR if that would be helpful (or to take over #45 if that's a thing that makes sense...)

mkborregaard:
It is, yes. See also https://github.com/JuliaPlots/StatPlots.jl/issues/72 where we're currently discussing this. It'd be great to resolve this now :-)

kescobo:
Ahh... I read that one, but didn't realize it was related but it seems clear now. Looks like ~ 4/7 currently open issues are related to this... good that the discussion is being had.

piever:
I'm also trying to come up with a solution for #44  (and this sort of group related issues). As noted already, the key of the problem is [here](https://github.com/JuliaPlots/Plots.jl/blob/master/src/series.jl#L513), where `group` is making the series. What it should do, instead, is to rearrange the `y` argument into a Matrix or Array of Arrays.

When the `groupedbar` recipe is being processed, the only available information at the time of grouping is the type of `args` (in this case it's a `GroupedBar`, with one field, itself called `args` containing the tuple of the arguments of the plot call). If we want to keep it a type recipe, the only solution that came to my mind is the following:

Add in `RecipesBase` a function:

`group_as_matrix(t) = false`

Then [here](https://github.com/JuliaPlots/Plots.jl/blob/master/src/series.jl#L513) it should check `group_as_matrix(args)` and if it's false, it groups normally with the series, otherwise it does a  Matrix. After that, to specify that we want the `groupedbar` to be grouped as a Matrix, we add a dispatch in StatPlots:

`RecipesBase.group_as_matrix(t::GroupedBar) = true`

mkborregaard:
Why would that function need to be in RecipesBase?

mkborregaard:
[...]

PR #192, opened 2018-10-10
Grouped box- and violin plots

kescobo:
Goal to address #145 . Testing with:

```
using Random
using DataFrames
using StatPlots

df = DataFrame(x=repeat([1,2,3],4), y=rand(12), g=vcat(fill("a", 6), fill("b", 6))
    )

@df df groupedbar(:x, :y, group=:g)
```
screen shot 2018-10-11 at 14 23 45

```
@df df boxplot(:x, :y, group=:g)
```
screen shot 2018-10-11 at 14 24 38

```
@df df violin(:x, :y, group=:g)
```
screen shot 2018-10-11 at 14 25 04

EDIT: this also addresses #56 and #89

### DIFF
--- a/src/boxplot.jl
+++ b/src/boxplot.jl
@@ -23,7 +23,7 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
     @assert whisker_width == :match || whisker_width >= 0 "whisker_width must be :match or a positive number"
     ww = whisker_width == :match ? bw : whisker_width
 
-    plotattributes[:group] == nothing ? ngroups = 1 : ngroups = length(unique(plotattributes[:group]))
+    isa(plotattributes[:group], Nothing) ? ngroups = 1 : ngroups = length(unique(plotattributes[:group]))
     thisgroup = plotattributes[:series_plotindex]
 
     outercenter = 0.5 * bw - bw / 2ngroups
--- a/src/boxplot.jl
+++ b/src/boxplot.jl
@@ -23,7 +23,7 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
     @assert whisker_width == :match || whisker_width >= 0 "whisker_width must be :match or a positive number"
     ww = whisker_width == :match ? bw : whisker_width
 
-    ngroups = length(unique(plotattributes[:group]))
+    plotattributes[:group] == nothing ? ngroups = 1 : ngroups = length(unique(plotattributes[:group]))
     thisgroup = plotattributes[:series_plotindex]
 
     outercenter = 0.5 * bw - bw / 2ngroups
--- a/src/boxplot.jl
+++ b/src/boxplot.jl
@@ -21,6 +21,24 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
     bw == nothing && (bw = 0.8)
     @assert whisker_width == :match || whisker_width >= 0 "whisker_width must be :match or a positive number"
     ww = whisker_width == :match ? bw : whisker_width
+
+    nr, nc = size(y)
+    x = if nc == 1
+        x
+    else
+        bws = bw / nc
+        bar_width := bws
+        xmat = zeros(nr,nc)
+        for r=1:nr
+            loopbw = _cycle(bws, r)
+            farleft = xnums[r] - 0.5 * (loopbw * nc)
+            for c=1:nc
+                xmat[r,c] = farleft + 0.5loopbw + (c-1)*loopbw
+            end
+        end
+        xmat
+    end
+
     for (i,glabel) in enumerate(glabels)
         # filter y
         values = y[filter(i -> _cycle(x,i) == glabel, 1:length(y))]
--- a/src/boxplot.jl
+++ b/src/boxplot.jl
@@ -6,6 +6,7 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
 
 @recipe function f(::Type{Val{:boxplot}}, x, y, z; notch=false, range=1.5, outliers=true, whisker_width=:match)
     # if only y is provided, then x will be UnitRange 1:length(y)
+
     if typeof(x) <: AbstractRange
         if step(x) == first(x) == 1
             x = plotattributes[:series_plotindex]
@@ -22,28 +23,20 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
     @assert whisker_width == :match || whisker_width >= 0 "whisker_width must be :match or a positive number"
     ww = whisker_width == :match ? bw : whisker_width
 
-    nr, nc = size(y) # size(y) == (6,)
+    ngroups = length(unique(plotattributes[:group]))
+    thisgroup = plotattributes[:series_plotindex]
 
-    x = if nc == 1
-        x
-    else
-        bws = bw / nc
-        bar_width := bws
-        xmat = zeros(nr,nc)
-        for r=1:nr
-            loopbw = _cycle(bws, r)
-            farleft = xnums[r] - 0.5 * (loopbw * nc)
-            for c=1:nc
-                xmat[r,c] = farleft + 0.5loopbw + (c-1)*loopbw
-            end
-        end
-        xmat
-    end
+    outercenter = 0.5 * bw - bw / 2ngroups
+    offsets = collect(-outercenter:(bw / ngroups):outercenter)
 
+    @info "offsets" offsets
+    # x = x .+ offsets[thisgroup]
+
+    @info "x" x
     for (i,glabel) in enumerate(glabels)
         # filter y
         values = y[filter(i -> _cycle(x,i) == glabel, 1:length(y))]
-
+        @info "y" y
         # compute quantiles
         q1,q2,q3,q4,q5 = quantile(values, Base.range(0,stop=1,length=5))
 
@@ -58,6 +51,7 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
 
         # make the shape
         center = Plots.discrete_value!(plotattributes[:subplot][:xaxis], glabel)[1]
+
         hw = 0.5_cycle(bw, i) # Box width
         HW = 0.5_cycle(ww, i) # Whisker width
         l, m, r = center - hw, center, center + hw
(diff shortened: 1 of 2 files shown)


src/boxplot.jl
```diff
@@ -21,6 +21,25 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
     bw == nothing && (bw = 0.8)
     @assert whisker_width == :match || whisker_width >= 0 "whisker_width must be :match or a positive number"
     ww = whisker_width == :match ? bw : whisker_width
+
+    nr, nc = size(y) # size(y) == (6,)
```
kescobo: I copied this code from box.jl, but it's giving me an error at the moment because the return is a 1-element tuple.

I tried adding in the `grouped_xy()` function from `bar.jl` as well, but that doesn't seem to work. I'm guessing the `x` and `y` pulled in by `g.args...` in:

```
@recipe function f(g::GroupedBar)
    x, y = grouped_xy(g.args...)
     ...
```

Is different than from `@recipe function f(::Type{Val{:boxplot}}, x, y, z;...)`, but I don't understand enough to know how to fix.

mkborregaard:
Can you post the image?

kescobo:
@mkborregaard Which image? The ones from the current default?

mkborregaard:
Yes :-)

kescobo:
Done! I think I just need to know how to get the number of groups from the `plot_attributes`?

kescobo:
I'm a bit lost here - I just hard-coded in 2 groups for now, but I can't figure out how this would work. It looks like the series are being generated separately? So even if I can find the number of groups, I don't know how I would write code to use the group number to calculate an offset.

Eg, if I do:

```
    ...
    loops = 0
    for (i,glabel) in enumerate(glabels)
        loops += 1
        @info loops
    ...
```
The output is
```
[ Info: 1
[ Info: 2
[ Info: 3
[ Info: 1
[ Info: 2
[ Info: 3
```

mkborregaard:
It's been so long I touched the grouping machinery I don't remember it anymore. You should be able to get it directly with `length(unique(g))`, though, right?

kescobo:
Woo! 

```
@df df boxplot(:x, :y, group=:g)
```
screen shot 2018-10-15 at 11 02 05


and

```
@df df boxplot(:g, :y, group=:x)
```
screen shot 2018-10-15 at 11 02 17

kescobo:
Still not done - working on violin now

kescobo:
screen shot 2018-10-15 at 11 07 26


screen shot 2018-10-15 at 11 07 44

kescobo:
Is this what we expect:

```
@df df boxplot(:x, :y, group=:g, bar_width=0.4)
@df df violin(:x, :y, group=:g, bar_width=0.4)
```

screen shot 2018-10-15 at 11 10 05

screen shot 2018-10-15 at 11 10 11

Is there a way to add space between the bars that I'm missing? If not, I think this is ready (other than tests, which I have zero idea how to write).

piever:
I think tests on StatPlots plots were moved to Plots and then disabled, so for the time being there's no need to worry so much about them. Here the space from one `x` value to the next seems a bit excessive though, I personally prefer the one from the image here https://github.com/JuliaPlots/StatPlots.jl/pull/192#issuecomment-429890341

To add space for the different groups (say from the blue and orange) you can cheat a bit and reduce the `bw` of an additional say 20% (it's how we do for bar plots as far as I remember: `bw = 0.8/bw`). Though 20% is the distance across separate "chunks" so maybe this one could be a bit less extreme, maybe 10% (or `sqrt(20)`% if somehow a geometric progression is more pleasant to look at). At the same time here it looks like the space across chunks is much bigger than 20% of the chunk, not sure how that changed.

kescobo:
>I personally prefer the one from the image here #192 (comment)

Yeah - that's the default. The latter is with `bar_width` set to a smaller amount.

src/boxplot.jl
```diff
@@ -21,10 +22,20 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
     bw == nothing && (bw = 0.8)
     @assert whisker_width == :match || whisker_width >= 0 "whisker_width must be :match or a positive number"
     ww = whisker_width == :match ? bw : whisker_width
+
+    ngroups = length(unique(plotattributes[:group]))
+    thisgroup = plotattributes[:series_plotindex]
+
+    outercenter = 0.5 * bw - bw / 2ngroups
+    offsets = collect(-outercenter:(bw / ngroups):outercenter)
+    bw = bw / ngroups
```
piever: Maybe consider:

```julia
bw = bw / ngroups * 0.9
ww = ww / ngroups * 0.9
```

to allow for some space within boxes of the same group.

kescobo: I think having the bars touching (as in `groupedbar`) makes sense - I was just wondering if there would be an appropriate keyword to grab to adjust that.

FWIW, this is what your suggestion looks like:

screen shot 2018-10-15 at 15 05 10
screen shot 2018-10-15 at 15 05 00

kescobo:
Actually, hold off on this - the `scatter` for the outliers isn't adjusted yet, and I've got some stray logging messages

kescobo:
Actually, it's ok I think, but there's something weird about how colors are selected for the scatter. But that's a separate issue as far as I can tell. New example code:

```
using Random
using DataFrames
using StatPlots

function rand_outlier(magnitude::Int=3)
    v = rand()
    return rand([-magnitude, magnitude]) * v
end

function grouped_df(n::Int, nx::Int, ng::Int, ol::Int)
    return DataFrame(x=repeat(collect(1:nx), ng*n),
               y=[rand() < ol ? rand_outlier() : rand() for _ in 1:n*ng*nx],
               g=vcat([fill(randstring(), nx*n) for _ in 1:ng]...))
end
```

```
@df grouped_df(100, 3, 2, 0.1) boxplot(:x, :y, group=:g, legend=false)
```

image


```
@df grouped_df(100, 3, 1, 0.1) boxplot(:x, :y, group=:g, legend=false)
```

image


```
@df grouped_df(100, 5, 3, 0.1) boxplot(:x, :y, group=:g, legend=false, color = [:red :green :blue])
```

image

kescobo:
@piever @mkborregaard I think this is ready? Opening a separate issue for the outlier thing.

kescobo:
Uggggh... hold on. Now it doesn't work _without_ groups. 

```
julia> @df grouped_df(100, 3, 1, 0.1) boxplot(:x, :y, legend=false)
ERROR: MethodError: no method matching iterate(::Nothing)
Closest candidates are:
  iterate(::Core.SimpleVector) at essentials.jl:578
  iterate(::Core.SimpleVector, ::Any) at essentials.jl:578
  iterate(::ExponentialBackOff) at error.jl:171
  ...
```

kescobo:
OK, ready for real this time

mkborregaard:
Hi this is awesome - but how do you get a space between the boxes in a group? To me it looks weird that they touch.

src/boxplot.jl
piever: The last thing is whether we should get an attribute to avoid "touching" boxplots - and what the default should be. Maybe they can default with no distance and `boxplot(...., space = 0.1)` could create some space between them (like in the example above, with `bw = bw / ngroups * (1-space)`.

@mkborregaard: what do you think is a sensible name for the attribute and a sensible default?

piever:
I commented above on the "touching". I agree this is awesome btw, it fixes 1/3 of the issues of StatPlots :)

kescobo:
>how do you get a space between the boxes in a group? To me it looks weird that they touch.

We could hardcode something ([as I did here](https://github.com/JuliaPlots/StatPlots.jl/pull/192#discussion_r225282294)), but I'm not sure what keyword would make sense to specify this. At the moment, the bars in `groupedbar` are touching, so that's how I coded it. 

FWIW - that's how it works in `ggplot2` as well: https://www.r-graph-gallery.com/265-grouped-boxplot-with-ggplot2/

piever:
Given that `groupedbar` is touching and in `ggplot2` it's touching as well, I think this could be merged as is (touching) and we can later think what's a good keyword to specify the distance. It should probably be done in a separate PR as the same keyword should, IMO, also work for `groupedbar`

kescobo:
@piever Sounds good to me. I don't have anything to add at this point

kescobo:
I added an issue to keep track of that idea, but I think this PR is gtg :shipit:

src/boxplot.jl
mkborregaard: `spacing`? I think sth like 0.1 should be the default.

mkborregaard:
Great, thanks a lot @kescobo !

Ibrahim-a-Ibrahim:
> Woo!
> 
> ```
> @df df boxplot(:x, :y, group=:g)
> ```
> 
> screen shot 2018-10-15 at 11 02 05
> 
> and
> 
> ```
> @df df boxplot(:g, :y, group=:x)
> ```
> 
> screen shot 2018-10-15 at 11 02 17

I tried the following code and it sounds that it does not result any grouping.
```
using Plots
using StatsPlots

x = rand(2,24,3)
x1 = x[:,:,1]
```
groupedboxplot(x, bar_position = :dodge, bar_width=0.7)
groupedboxplot(x1, bar_position = :dodge, bar_width=0.7)
Any suggestion to end up with a grouped boxplot?

mkborregaard:
StatsPlots expect data to be "long" format, e.g. https://r4ds.had.co.nz/tidy-data.html

Ibrahim-a-Ibrahim:
> StatsPlots expect data to be "long" format, e.g. https://r4ds.had.co.nz/tidy-data.html

Could you please help to provide a piece of code to represent the issue that I raised it up using the "long" format data frame?

mkborregaard:
What do you expect the groups to be? You data has no categorical variables to group on

Ibrahim-a-Ibrahim:
> What do you expect the groups to be? You data has no categorical variables to group on

I am looking to plot the data from the 3D array like the attached photo. Is it helpful?
3phase

mkborregaard:
Not really - you are taking the `group` information from the column it is in, but the `x` variable is also not a grouped variable. You are looking for something like

```julia
using StatsPlots

x = rand(1:2, 100)
gr = rand(1:3, 100)
y = rand(100)

groupedboxplot(x, y, group = gr)
```

Skærmbillede 2020-08-19 kl  12 44 11

Although in this case, `groupedboxplot` probably should not be using numeric ticks.

kescobo:
Also - doesn't really make sense to have this discussion in a merged PR. Slack, zulip, or discourse are probably more appropriate.

kescobo:
Ah, I see you did that [here](https://discourse.julialang.org/t/grouped-boxplot-https-github-com-juliaplots-statsplots-jl-pull-192/44655/7?u=kevbonham), extra doesn't make sense to renew this old PR then (which was superceded by #211 in any case)

PR #204, opened 2019-01-20
slightly decrease size in cornerplot

mkborregaard:
Maybe this should be thought more through - I'm rather unhappy with both the size and transparency of points in cornerplot


mkborregaard:
Also changes the default marker colors in corner

PR #211, opened 2019-02-12
WIP: implement groupedboxplot and groupedviolin recipes

daschw:
My quick take on a groupedboxplot and groupedviolin recipe. I hope that I did not miss anything important.

```julia
using StatsPlots, DataFrames

df = DataFrame(x = repeat(["A", "B"], inner = 10), y = (1:20) .+ randn(20), g = repeat(["Group 1", "Group 2"], inner = 5, outer = 2))

@df df groupedviolin(:x, :y, group = :g)
```

violin

```julia
@df df groupedboxplot(:x, :y, group = :g)
```

boxplot

cf. https://github.com/JuliaPlots/StatsPlots.jl/issues/198, https://github.com/JuliaPlots/StatsPlots.jl/pull/192

### DIFF
--- a/src/bar.jl
+++ b/src/bar.jl
@@ -7,7 +7,7 @@ Plots.group_as_matrix(g::GroupedBar) = true
 grouped_xy(x::AbstractVector, y::AbstractArray) = x, y
 grouped_xy(y::AbstractArray) = 1:size(y,1), y
 
-@recipe function f(g::GroupedBar; spacing = 0)
+@recipe function f(g::GroupedBar; spacing = 0.1)
     x, y = grouped_xy(g.args...)
 
     nr, nc = size(y)
--- a/src/bar.jl
+++ b/src/bar.jl
@@ -4,8 +4,8 @@ recipetype(::Val{:groupedbar}, args...) = GroupedBar(args)
 
 Plots.group_as_matrix(g::GroupedBar) = true
 
-grouped_xy(x::AbstractVector, y::AbstractMatrix) = x, y
-grouped_xy(y::AbstractMatrix) = 1:size(y,1), y
+grouped_xy(x::AbstractVector, y::AbstractArray) = x, y
+grouped_xy(y::AbstractArray) = 1:size(y,1), y
 
 @recipe function f(g::GroupedBar)
     x, y = grouped_xy(g.args...)
--- a/src/bar.jl
+++ b/src/bar.jl
@@ -7,7 +7,7 @@ Plots.group_as_matrix(g::GroupedBar) = true
 grouped_xy(x::AbstractVector, y::AbstractArray) = x, y
 grouped_xy(y::AbstractArray) = 1:size(y,1), y
 
-@recipe function f(g::GroupedBar)
+@recipe function f(g::GroupedBar; spacing = 0)
     x, y = grouped_xy(g.args...)
 
     nr, nc = size(y)
@@ -32,7 +32,7 @@ grouped_xy(y::AbstractArray) = 1:size(y,1), y
         x
     else
         bws = plotattributes[:bar_width] / nc
-        bar_width := bws
+        bar_width := bws * clamp(1 - spacing, 0, 1)
         xmat = zeros(nr,nc)
         for r=1:nr
             bw = _cycle(bws, r)
--- a/src/bar.jl
+++ b/src/bar.jl
@@ -7,7 +7,7 @@ Plots.group_as_matrix(g::GroupedBar) = true
 grouped_xy(x::AbstractVector, y::AbstractArray) = x, y
 grouped_xy(y::AbstractArray) = 1:size(y,1), y
 
-@recipe function f(g::GroupedBar; spacing = 0.1)
+@recipe function f(g::GroupedBar; spacing = 0)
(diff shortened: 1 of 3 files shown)


mkborregaard:
That's really awesome! What do you think about creating a little spacing, as discussed on the original PR?

daschw:
Yes, that would be a nice feature. I'll look into it.

daschw:
Here's a pretty straight forward implementation of a spacing kw_arg for groupedbar, groupedboxplot and groupedviolin. It's set in relative terms between 0 and 1 and I made it default to 0.1. Not sure, if this is what you were looking for and if it's general enough, or if it would be better implemented as a Plots attribute.

bar_spacing
violin
boxplot
boxplot_spacing

I'm also open for other ideas or suggestions.

mkborregaard:
I think this is great, and ready to merge. Awesome work @daschw !
@piever would you like to review first?

piever:
Looks really good!

> @piever would you like to review first?

Yes, I'll review it tomorrow. I have one basic confusion though: why doesn't this need `group_as_matrix(::GroupedBoxplot) = true`? I thought that otherwise [here](https://github.com/JuliaPlots/Plots.jl/blob/42aa7d8d55e872c105a45c70dc7b2e9bef6f00e6/src/series.jl#L636) it would already pass to the recipe the "split subdatasets", instead it seems to just pass along the "whole" vectors.

src/boxplot.jl
```diff
@@ -118,3 +118,47 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
...
+
+recipetype(::Val{:groupedboxplot}, args...) = GroupedBoxplot(args)
+
+@recipe function f(g::GroupedBoxplot; spacing = 0.1)
+    x, y = grouped_xy(g.args...)
+
+    # extract xnums and set default bar width.
+    # might need to set xticks as well
+    x = if eltype(x) <: Number
+		bar_width --> (0.8 * mean(diff(x)))
```
piever: Indentation is a bit strange here and in the rest of the diff.

src/boxplot.jl
```diff
@@ -118,3 +118,47 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
...
+        float.(x)
+    else
+		bar_width --> 0.8
+        ux = unique(x)
```
piever: I'm a bit unsure what's going on here in terms of when the "grouping" is done. How can this sidestep the machinery in https://github.com/JuliaPlots/Plots.jl/blob/399a9b16f6b4a9259c5c844b38954bc56715a1ba/src/args.jl#L1062 and https://github.com/JuliaPlots/Plots.jl/blob/42aa7d8d55e872c105a45c70dc7b2e9bef6f00e6/src/series.jl#L634?

src/bar.jl
```diff
@@ -4,10 +4,10 @@ recipetype(::Val{:groupedbar}, args...) = GroupedBar(args)
 
 Plots.group_as_matrix(g::GroupedBar) = true
 
-grouped_xy(x::AbstractVector, y::AbstractMatrix) = x, y
-grouped_xy(y::AbstractMatrix) = 1:size(y,1), y
+grouped_xy(x::AbstractVector, y::AbstractArray) = x, y
+grouped_xy(y::AbstractArray) = 1:size(y,1), y
 
-@recipe function f(g::GroupedBar)
+@recipe function f(g::GroupedBar; spacing = 0.1)
```
piever: Like the name `spacing`! Somehow I feel it should default to `0` to preserve the old behavior of `groupedbar`. I wonder if there is a way to make it clear that we refer to the spacing "within groups" whereas the ration "spacing between groups / spacing within groups" IIUC is hardcoded.

src/boxplot.jl
daschw: I think the user recipe is applied before the grouping recipe, no?

src/bar.jl
daschw: I've changed the default spacing of `groupedbar` to `0` again. Is it ok toleave it at `0.1` for `groupedboxplot` and `groupedviolin`? The spacing between groups can be set with the `bar_width` attribute.

src/boxplot.jl
daschw: hopefully fixed

daschw:
> why doesn't this need `group_as_matrix(::GroupedBoxplot) = true`? I thought that otherwise [here](https://github.com/JuliaPlots/Plots.jl/blob/42aa7d8d55e872c105a45c70dc7b2e9bef6f00e6/src/series.jl#L636) it would already pass to the recipe the "split subdatasets", instead it seems to just pass along the "whole" vectors.

I think this happens after the user recipe is applied.
Also, `group_as_matrix` only allows one value for each (x, group) combination, which makes no sense for boxplot or violin.

piever:
> I think this happens after the user recipe is applied. Also, `group_as_matrix` only allows one value for each (x, group) combination, which makes no sense for boxplot or violin.

Fair point, the same approach as `groupedbar` would actually not work here.

mkborregaard:
I feel like it would be nice to have this described in the readme for discovery?

PR #215, opened 2019-02-18
Optionally show mean, median and quantiles in violin (Fix #127)

### LINKED ISSUE #127: Violin with mean
It would be nice to have an option of adding a mean/median line to the violin plots. These lines should be "halved" if `side` is not equal to `:both`.

daschw:
This adds the kwargs `mean`, `median` and `quantiles` to the violin recipe. `mean` and `median` accept Bools. `quantiles` accepts a vector of quantiles or an integer for evenly spaced quantiles. Quantiles are shown as a solid line and the mean is plotted as a dotted line. I'm open for other ideas and suggestions.

cc: @yakir12

```julia
using StatsPlots
y = [i * randn(100) for i in 1:4]
```
```julia
violin(y, median = true)
```
violin1

```julia
violin(y, quantiles = [0.1, 0.5, 0.9], linecolor = :white, linewidth = 3)
```
violin2

```julia
violin(y, quantiles = 5, trim = false, linecolor = :match, fillalpha = 0.3)
```
violin3

```julia
violin(y, quantiles = 3,  mean = true)
```
violin4

### DIFF
--- a/src/violin.jl
+++ b/src/violin.jl
@@ -19,7 +19,7 @@ get_quantiles(x::Real) = [x]
 get_quantiles(b::Bool) = b ? [0.5] : Float64[]
 get_quantiles(n::Int) = range(0, 1, length = n + 2)[2:end-1]
 
-@recipe function f(::Type{Val{:violin}}, x, y, z; trim=true, side=:both, median = false, quantiles = Float64[])
+@recipe function f(::Type{Val{:violin}}, x, y, z; trim=true, side=:both, mean = false, median = false, quantiles = Float64[])
     # if only y is provided, then x will be UnitRange 1:length(y)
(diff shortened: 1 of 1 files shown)


yakir12:
Wow, cool!!!

mkborregaard:
Huh, this is interesting functionality. Never seen it before but it looks useful.

daschw:
I saw something similar here: https://matplotlib.org/gallery/statistics/boxplot_vs_violin.html

daschw:
Maybe `showmeans` and `showmedians` would be the better kwarg names?

piever:
I use `show_median` in StatsMakie: I think it's a bit clearer than just `median`.

PR #216, opened 2019-02-24
Simplify interface and port groupedviolin/boxplot to standard plots grouping

piever:
This started as a simplification of the UI code (now that it is possible to use Interact widgets cleanly without depending on Interact). Halfway through I noticed that there were issues incorporating `groupedboxplot` and `groupedviolin` because, as they reimplement the grouping mechanism, they miss some cases, i.e. when `group = (vec1, vec2)` (grouping by more than one thing) and `group = (name1=vec1, name2=vec2)` (named grouping, when the legend becomes `name1 = val1, name2 = val2`) so I ported them to use the `extractGroupArgs` machinery.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -9,7 +9,7 @@ IterableTables 0.5
 TableTraitsUtils 0.1
 TableTraits
 DataValues
-Widgets 0.4.0
+Widgets 0.5.0
 Observables 0.2.2
 DataStructures
 Clustering
--- a/src/boxplot.jl
+++ b/src/boxplot.jl
@@ -146,12 +146,13 @@ recipetype(::Val{:groupedboxplot}, args...) = GroupedBoxplot(args)
     # shift x values for each group
     group = get(plotattributes, :group, nothing)
     if group != nothing
-        ug = unique(group)
-        n = length(ug)
+        gb = Plots.extractGroupArgs(group)
+        labels, idxs = getfield(gb, 1), getfield(gb, 2)
+        n = length(labels)
         bws = plotattributes[:bar_width] / n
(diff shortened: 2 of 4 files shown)


piever:
Now this becomes possible:

```julia
@df school groupedviolin(:Minrty, :SSS, group = {:Sx, :Sector})
```

screenshot from 2019-02-24 12-01-25

mkborregaard:
Nice! - what's that syntax with the curly brackets?

piever:
> Nice! - what's that syntax with the curly brackets?

It's one of the `@df` features (taken from Query): `{:Sx, :Sector}` becomes `(Sx =:Sx, Sector = :Sector)` so that the user doesn't need to type in the name.
The more explicit command `@df school groupedviolin(:Minrty, :SSS, group = (Sx =:Sx, Sector = :Sector))` would also work and allows legend names different from the column names.

mkborregaard:
Nice - and I guess `(:Sx, :Sector)` couldn't just expand to that automatically?

piever:
> Nice - and I guess `group = (:Sx, :Sector)` couldn't just expand to that automatically?

Not in an obvious way (I don't like the idea of the macro looking for a specific keyword in the function call). To remove `{}` syntax in favor of normal Julia syntax for NamedTuples I think the best way would be something like [this comment](https://github.com/JuliaLang/julia/issues/28579#issuecomment-412950407) to allow `(; a, b)` to mean `(a=a, b=b)` in Base Julia (there is an open issue for it that I couldn't find).

mkborregaard:
Yes, it was the keyword search I meant but I can follow your dislike of it.

daschw:
Awesome!

PR #226, opened 2019-04-17
Added dotplot - show all datapoints for a category

BioTurboNick:
There's a trend in science to show all individual datapoints when something like a boxplot would be used. This dotplot is designed to allow individual data points to be overlaid onto a boxplot. Violin plots, while showing the overall shape of a distribution, may be deceptive for sparse points.

Currently, it uses a window around each point (based on quantiles) and randomization to spread the points out to keep the point density the same across the plot. May be extended later to other forms (no-overlap).

```
using StatsPlots
x = mod.(rand(Int, 5000), 5) .+ 1
y = randn(5000)
dotplot(x,y, marker=(3, stroke(0)))
boxplot!(x,y, opacity=0.5)
```

dotplot1

A future addition could allow raincloud plots to be created: https://micahallen.org/2018/03/15/introducing-raincloud-plots/


BioTurboNick:
Added a `widthwindow` argument that allows the area over which density is assessed to change. Default is 0.1 of the distance between q1 and q5, valid range [0,1]. Value of 1 creates a uniform width scatter.

src/dotplot.jl
```diff
@@ -0,0 +1,91 @@
...
+    points_x, points_y = zeros(0), zeros(0)
+    bw = plotattributes[:bar_width]
+    bw == nothing && (bw = 0.8)
+    0.0 ≤ widthwindow ≤ 1.0 || throw(ArgumentError("$(:widthwindow) must be in the range [0,1]"))
+    for (i,glabel) in enumerate(glabels)
+        # filter y
+        values = y[filter(i -> _cycle(x,i) == glabel, 1:length(y))]
+
+        # compute quantiles
+        q1,q2,q3,q4,q5 = quantile(values, Base.range(0,stop=1,length=5))
```
daschw: I think you don't really need 5 quantiles here.
```julia
vmin, vmax = extrema(values)
nearby = widthwindow * (vmax - vmin)
```

src/dotplot.jl
```diff
@@ -0,0 +1,91 @@
...
+
+        # make the shape
+        center = Plots.discrete_value!(plotattributes[:subplot][:xaxis], glabel)[1]
+        hw = 0.5_cycle(bw, i) # Box width
+
+        nearby = widthwindow * abs(q5 - q1)
+        countnear = [count((values .< (x + nearby)) .& (values .> (x - nearby))) for x ∈ values]
+
+        pw = hw / maximum(countnear)
+        offsets = [(rand() * 2 - 1) * pw * x for x ∈ countnear]
```
daschw: This means that plots with the same data will be different each time, right?
Do you think there is also a unique method of distributing the points vertically?

piever:
> However, I'm not quite sure about the name.

I'm not an expert on the `violin`, `boxplot` & co. nomenclature, but isn't this very similar to the beeswarm plot from https://github.com/JuliaPlots/StatsPlots.jl/pull/61? Maybe it could also be called beeswarm. I wonder whether this could have a `side` attribute (pretty much like violin) to show points only on one side, so that users could combine it with a violin on the other side and achieve the same plots as in https://github.com/JuliaPlots/StatsPlots.jl/pull/61#issuecomment-318675054

> I'm also not sure if it is a problem that plots look different for the same data due to the random horizontal distribution of the points. Again, I can't think of a better implementation.

I think ggplot2 in R also has some jitter option, maybe we could check how they do it (if they accept that the same data could give different plots, or fix some random seed or some other solution I can't think of).

BioTurboNick:
Ahh, that's the name of that plot, `beeswarm`. I had no idea and so couldn't easily find it. Take that for what it's worth :-). Searching `dotplot` for images shows everything but the beeswarm layout.

#61 is definitely going beyond my immediate goals, but I do think the efforts should be combined. I'll see how I can help; perhaps a simpler version using jitter alone (seems to be the term in use) can be added as a step to the more complex one. Being able to specify (or retrieve?) a seed might be useful in case one finds a horizontal distribution they like...

mkborregaard:
Thanks for this @BioTurboNick . So, I think this is a beeswarm plot + a boxplot. I think it would definitely be nice to combine this with #61 - the code I posted in a comment creates a beeswarm plot which I personally find superior to any other implementation I've seen. It does have some spikes, but that is the result from wanting to show the dots like this.

As for this, I think the ideomatic way to do this in StatsPlots would be 
```
beeswarm(myx, myy)
boxplot!(myx, myy)
```
We could have a recipe that combines these - in that case IMHO it should be named for the components, eg like `scatterlines`, so `beebox` or sth like that. 

@BioTurboNick would you consider trying to merge your beeswarm implementation with the implementation I posted? Or do you actually prefer the jittered version?

BioTurboNick:
@mkborregaard - Yeah, I'd be down to tackle that. It'll be a few weeks, trying to get a paper out and have a vacation.

BioTurboNick:
I've looked at the Wilkinson paper on dot plots and I'm persuaded that "dot plot" is the proper general term for this type of plot.

I added a non-displaced version and a more refined jittered version based on the violin plot. `mode` argument set to `:none` (default) is the first; `:densityjittered` for my original; `violinjittered` for the violin version. This is mainly experimental use.

I think the version based on violin is better.

src/dotplot.jl
```diff
@@ -0,0 +1,183 @@
...
+    if typeof(x) <: AbstractRange
+        if step(x) == first(x) == 1
+            x = plotattributes[:series_plotindex]
+        else
+            x = [getindex(x, plotattributes[:series_plotindex])]
+        end
+
+        x = repeat([x], length(y))
+    end
+    x = Float64.(x)
```
daschw:
```julia
df = DataFrame(x = repeat(["A", "B"], outer = 100), y = randn(200))
@df df dotplot(:x, :y)
```
errors here:
```
ERROR: MethodError: no method matching Float64(::String)
Closest candidates are:
  Float64(::Int8) at float.jl:60
  Float64(::Int16) at float.jl:60
  Float64(::Int32) at float.jl:60
  ...
Stacktrace:
 [1] _broadcast_getindex_evalf at .\broadcast.jl:578 [inlined]
 [2] _broadcast_getindex at .\broadcast.jl:551 [inlined]
 [3] getindex at .\broadcast.jl:511 [inlined]
 [4] copy at .\broadcast.jl:787 [inlined]
 [5] materialize(::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1},Nothing,Type{Float64},Tuple{Array{String,1}}}) at .\broadcast.jl:753
 [6] macro expansion at C:\Users\Daniel\.julia\dev\StatsPlots\src\dotplot.jl:16 [inlined]
 [7] apply_recipe(::Dict{Symbol,Any}, ::Type{Val{:dotplot}}, ::Array{String,1}, ::Array{Float64,1}, ::Nothing) at C:\Users\Daniel\.julia\packages\RecipesBase\OPemU\src\RecipesBase.jl:275
 [8] _process_seriesrecipe(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}) at C:\Users\Daniel\.julia\dev\Plots\src\pipeline.jl:408
 [9] _plot!(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Tuple{Array{String,1},Array{Float64,1}}) at C:\Users\Daniel\.julia\dev\Plots\src\plot.jl:234
 [10] #plot#133(::Base.Iterators.Pairs{Symbol,Symbol,Tuple{Symbol},NamedTuple{(:seriestype,),Tuple{Symbol}}}, ::Function, ::Array{String,1}, ::Vararg{Any,N} where N) at C:\Users\Daniel\.julia\dev\Plots\src\plot.jl:57
 [11] (::getfield(RecipesBase, Symbol("#kw##plot")))(::NamedTuple{(:seriestype,),Tuple{Symbol}}, ::typeof(plot), ::Array{String,1}, ::Array{Float64,1}) at .\none:0
 [12] #dotplot#103(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Array{String,1}, ::Vararg{Any,N} where N) at C:\Users\Daniel\.julia\packages\RecipesBase\OPemU\src\RecipesBase.jl:369
 [13] dotplot(::Array{String,1}, ::Vararg{Any,N} where N) at C:\Users\Daniel\.julia\packages\RecipesBase\OPemU\src\RecipesBase.jl:369
 [14] #add_label#17(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Array{String,1}, ::Function, ::Array{String,1},
::Vararg{Any,N} where N) at C:\Users\Daniel\.julia\dev\StatsPlots\src\df.jl:139
 [15] add_label(::Array{String,1}, ::Function, ::Array{String,1}, ::Vararg{Any,N} where N) at C:\Users\Daniel\.julia\dev\StatsPlots\src\df.jl:137
 [16] (::getfield(Main, Symbol("##51#52")))(::DataFrame) at .\none:0
 [17] top-level scope at none:0
```
`@df df violin(:x, :y)` works, if you want to look that recipe is implemented.

BioTurboNick: Yeah, the cause was a quick hack to get around something I saw. Thanks for pointing out it is a problem for df.

BioTurboNick:
I had a question about the code at the top for when only y is provided. It's taken from `boxplot`. The comment says `if only y is provided, then x will be UnitRange 1:length(y)`. The seems to be incorrect or incomplete.

The behavior of the code appears to be that if `y` is a 2-dimensional vector, x will be `UnitRange 1:size(y, 2)`; if `y` is a vector of vectors, x will be `UnitRange 1:length(y)`; if `y` is a vector of numbers, x will be just 1. Can we adjust this comment to match?

daschw:
> Can we adjust this comment to match?

Sure, feel free to add a better comment.

BioTurboNick:
Here's a figure comparing the three modes, now specified by `mode = none`, `mode = :uniform`, and `mode = :density`.

compare

Thinking through the possible variants, we could have:
- None
- Uniform with jitter
- Uniform with no overlap (not yet implemented)
- Kernel density with jitter
- Kernel density with no overlap (not yet implemented)
- Beeswarm (from #61 proposal)

If these are all implemented in this one recipe, how should they be accessed?

Could do: `:none`, `:uniform`, `:density`, `:beeswarm`; uniform and density take additional argument of `jitter = true`.

Thoughts?

mkborregaard:
I like your thinking there

mkborregaard:
(the conflict is the REQUIRE file - maybe just delete that and add the dep to the Project.toml instead?

BioTurboNick:
Guess they conflict either way.

mkborregaard:
Yeah - do you know how to do a local rebase?

BioTurboNick:
@mkborregaard - I think so, if I know what the goal is?

mkborregaard:
The thing is that while this PR has been open, we've abandoned REQUIRE files and started using Project.toml files instead. But this PR points to an older commit on master, so when you add a project file it would overwrite the existing one. 
So by doing a local rebase, you would address all conflicts manually and have the PR point towards the current master. It's easiest to show in an image:
Skærmbillede 2019-06-04 kl  13 16 40

mkborregaard:
As you can see your groupspacingfix PR is also to the old master, which is what makes the commit tree look a bit non-linear.

BioTurboNick:
Ah, I see. I'll figure it out, thanks!

mkborregaard:
I can warmly recommend gitkraken instead of fiddling around with the command line. It's just rebase->resolve->force push.

BioTurboNick:
Oh awesome tool, thanks!

mkborregaard:
You don't have to commit the manifest file - just delete it. It's not necessary for packages.

PR #229, opened 2019-05-21
Plot recipe for MDS from MultivariateStats.jl

kescobo:
This is dependent on https://github.com/JuliaStats/MultivariateStats.jl/pull/85, though I'm not quite sure how to indicate that in compat until it's tagged. 

```
using StatsPlots
using Distances
using MultivariateStats

m = rand(100, 10)

dm = pairwise(BrayCurtis(), m)

mds = fit(MDS, dm, distances=true)

plot(mds)
```

mds

### DIFF
--- a/src/mds_scatter.jl
+++ b/src/mds_scatter.jl
@@ -1,0 +1,10 @@
+@recipe function f(mds::MDS{<:Real})
+    xticks := false
+    yticks := false
+    ev = eigvals(mds)
+    var_explained = [v / sum(ev) for v in ev]
+    xlabel --> "MDS1 ($(round(var_explained[1] * 100, digits = 2))%)"
+    ylabel --> "MDS2 ($(round(var_explained[2] * 100, digits = 2))%)"
+    seriestype := :scatter
+    projection(mds)[:,1], projection(mds)[:,2]
+end
--- a/src/mds_scatter.jl
+++ b/src/mds_scatter.jl
@@ -1,10 +1,22 @@
-@recipe function f(mds::MDS{<:Real})
-    xticks := false
-    yticks := false
+@recipe function f(mds::MDS{<:Real}; mds_axes=(1,2))
+    length(mds_axes) in [2,3] || throw(ArgumentError("Can only accept 2 or 3 mds axes"))
+    xax = mds_axes[1]
+    yax = mds_axes[2]
     ev = eigvals(mds)
     var_explained = [v / sum(ev) for v in ev]
-    xlabel --> "MDS1 ($(round(var_explained[1] * 100, digits = 2))%)"
-    ylabel --> "MDS2 ($(round(var_explained[2] * 100, digits = 2))%)"
+    proj = projection(mds)
+
+    xlabel --> "MDS$xax ($(round(var_explained[xax] * 100, digits = 2))%)"
+    ylabel --> "MDS$yax ($(round(var_explained[yax] * 100, digits = 2))%)"
     seriestype := :scatter
-    projection(mds)[:,1], projection(mds)[:,2]
+    xticks := false
+    yticks := false
+
+    if length(mds_axes) == 3
+        zax = mds_axes[3]
+        zlabel --> "MDS$zax ($(round(var_explained[zax] * 100, digits = 2))%)"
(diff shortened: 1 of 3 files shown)


mkborregaard:
That looks sweet! Have you considered an interface for multiple dimensions? Either passing dimensions manually, or using subplots or possibly 3d?

kescobo:
@mkborregaard I have considered, but wasn't sure about the implementation. Since this is not a user recipe, can I add a keyword for that?

mkborregaard:
Sure :-)

mkborregaard:
(technically this is a user recipe)

kescobo:
I think I was mis-remembering the deal with keyword arguments... 

In any case, this now allows a user to specify axes, and do a 3rd axis:

```julia
using StatsPlots
using Distances
using MultivariateStats

m = rand(100, 10)
dm = pairwise(BrayCurtis(), m)
mds = fit(MDS, dm, distances=true)

plot(mds)
```

mds1

```julia
plot(mds, mds_axes=(3,5))
```

mds2


```julia
plot(mds, mds_axes=(1,2,3))
```

mds3

mkborregaard:
Love it :heart:

mkborregaard:
Oh - I merged already, but only now realised - shouldn't `aspect_ratio` have been set to 1?

src/mds_scatter.jl
```diff
@@ -0,0 +1,22 @@
+@recipe function f(mds::MDS{<:Real}; mds_axes=(1,2))
```
mkborregaard: Also, I think you're missing a dep on MultivariateStats for this to work?

mkborregaard: Wait, you do have that, I seem to have a local problem, sorry for the noise

kescobo:
Not sure, what does that do? Happy to add it in.

mkborregaard:
@kescobo I'm getting `WARNING: could not import MultivariateStats.MDS into Main` locally - is there something wrong with my setup/versions?

Also, could the signature be wider to allow for PCAs and other ordinations as well? Is the interface not general enough for this?

It would be awesome to have a readme example with this code
```julia
using MultivariateStats, RDatasets, StatsPlots

Xtr = convert(Matrix, iris[:, 1:4])'
M = fit(PCA, Xtr; maxoutdim=2)
plot(M, group = iris.Species)
```
or something like that. Compare the code here: https://multivariatestatsjl.readthedocs.io/en/latest/pca.html#data-analysis

mkborregaard:
> Not sure, what does that do?

Forces the axes to have the same scale. Given that they are by definition on the same scale it gives a more correct representation of the spread on the axes.

kescobo:
>@kescobo I'm getting WARNING: could not import MultivariateStats.MDS into StatsPlots locally - is there something wrong with my setup/versions?

You have to be on master of MultivariatStats (see the first comment in this thread=). The PR introducing this API hasn't been tagged, so I can't add it to `[compat]` in the `Project.toml` yet (I asked on slack)

>Also, could the signature be wider to allow for PCAs and other ordinations as well? Is the interface not general enough for this?

I haven't looked into it, but I thought about the same thing. I will happily do a separate PR.

>It would be awesome to have a readme example with this code

Sure - should this be a separate PR since this one is already merged?

kescobo:
> Forces the axes to have the same scale. Given that they are by definition on the same scale. Given that they are by definition on the same scale it gives a more correct representation of the spread on the axes.

My understanding was that for PCoA at least, the scales are essentially unitless and so scaling is meaningless. Entirely possible I misunderstood that though 😆

mkborregaard:
> You have to be on master of MultivariatStats (see the first comment in this thread=)

Sorry :joy:

> I will happily do a separate PR.

Yay! 💖 

> should this be a separate PR

I think it has to be yeah.

> scales are essentially unitless 

AFAIK the PCoA transformation maintains distances among points. But if Plots chooses different scales for the x and y axis (based on the range of the data and the shape of the plotting window), the distances will not be correct.

kescobo:
>AFAIK the PCoA transformation maintains distances among points.

But PCoA is for non-metric distances right? Maybe I'll ask on Slack...

PR #231, opened 2019-05-27
Drop older versions of plots

kescobo:
See https://github.com/JuliaRegistries/General/pull/902#discussion_r287570036


kescobo:
cc @mkborregaard

mkborregaard:
Nice. @piever this seemed cleaner to me given the discussion on https://github.com/JuliaRegistries/General/pull/902
If you agree I'll just tag this as the new version.

Issue #234, opened 2019-06-04
Grouped Boxplot bar width broken with numeric X values


BioTurboNick:
```
a = randn(6000);
b = repeat([1;2;3], 2000);
c = repeat([1;2], 3000);
groupedboxplot(b,a,group=c)
```
This is producing infinitesimally thin boxes. It appears to be due to line 135 in boxplot.jl:
`bar_width --> (0.8 * mean(diff(x)))`

`mean(diff(x))` evaluates to ~0.0003334.

It looks like it was copied from groupedbar, which just has one x value per column.

BioTurboNick:
I believe `bar_width --> (0.8 * mean(diff(unique(x))))` achieves what was desired. I'll create a PR.

mkborregaard:
Thanks for closing this (and the PR). Pro tip: add "closes #234" to the title of the PR to have it close the issue automatically when merged.

PR #235, opened 2019-06-04
Fixed Grouped Boxplot default bar_width when X is Numeric

BioTurboNick:
Copied the code from `groupedviolin` to `groupedboxplot` Presumably some of this should be consolidated?

Also updated the comment at the top of violin and boxplot to reflect the true behavior when x is absent.

Addresses #234


mkborregaard:
Nice, thanks!

PR #237, opened 2019-06-06
replaced isnothing with === nothing for compatbility with Julia 1.0

torfjelde:
`isnothing(x)` was introduced in Julia 1.1 as an alias for `x === nothing`, but `isnothing` is not defined in Julia 1.0; this PR just fixes the resulting `isnothing` not defined error.


mkborregaard:
thanks

PR #253, opened 2019-08-15
Allow "arbitrary" arguments to corrplot

asinghvi17:
As of now, you can only supply matrices to corrplot; however, to my understanding, corrplot is applied relatively early in the plotting pipeline, and cannot be accessed through a seriestype.  That being said, it would be nice for user-defined types which are "`corrplot`-able" to be easily plottable through a function overload.

Therefore, I've introduced here a function `to_corrplot_matrix(mat)`, which can be overloaded by the user for their custom types.  It must return a matrix of a form which can be ingested by `corrplot`.

It would be great if there's a recipe-based solution to this that I'm missing; however, this kind of "multiple-dispatch" based recipe seems hard to implement without this kind of workaround (please tell me if I'm wrong here).  Perhaps it would be possible to introduce a plot-type-based type conversion recipe syntax, as well?

### DIFF
--- a/src/corrplot.jl
+++ b/src/corrplot.jl
@@ -2,6 +2,14 @@
 
 recipetype(::Val{:corrplot}, args...) = CorrPlot(args)
 
+"""
+    to_corrplot_matrix(mat)
+
+Transforms the input into a correlation plot matrix.  
+Meant to be overloaded by other types!
+"""
+to_corrplot_matrix(mat::AbstractMatrix) = mat
+
 function update_ticks_guides(d::KW, labs, i, j, n)
     # d[:title]  = (i==1 ? _cycle(labs,j) : "")
     # d[:xticks] = (i==n)
@@ -11,7 +19,7 @@ function update_ticks_guides(d::KW, labs, i, j, n)
 end
 
 @recipe function f(cp::CorrPlot)
-    mat = cp.args[1]
+    mat = to_corrplot_matrix(cp.args[1])
     n = size(mat,2)
     C = cor(mat)
     labs = pop!(plotattributes, :label, [""])


src/corrplot.jl
```diff
@@ -11,7 +19,7 @@ function update_ticks_guides(d::KW, labs, i, j, n)
 end
 
 @recipe function f(cp::CorrPlot)
-    mat = cp.args[1]
+    mat = to_corrplot_matrix(cp.args[1])
```
piever: Should we just mimic Makie and have a `convert_arguments(::Type{CorrPlot}, args...)` as a method to overload? I find it odd to use a function specific to `CorrPlot`.

piever:
The issue with what recipe happens when is what led to using a `@df` macro for `DataFrame`, otherwise the column expansion would happen after `corrplot`. What is the data you are trying to do a `corrplot` on? Is it a tabular format or something completely different?

> Perhaps it would be possible to introduce a plot-type-based type conversion recipe syntax, as well?

I like the idea, not sure how easy or hard is the implementation. Maybe the `@recipe` macro could run `args = convert_arguments(PlotType, args...)` by default?

asinghvi17:
That makes sense, it would probably dispatch on `Val(:seriestype)` or something - would need to make it recursive, though.  

> What is the data you are trying to do a corrplot on? Is it a tabular format or something completely different?

Unfortunately it's not in a tabular format, which was why I wanted to create this function.

With respect to convert_arguments: we could definitely do it, not sure where it should be placed, though.  Will look through the Plots code to understand it further.  What would be the best way to define these recipes?

src/corrplot.jl
asinghvi17: Agreed, that would be better - would probably require a substantial modification to the recipe system, though.

piever:
I imagine a unified `convert_arguments` could be added by the [recipe](https://github.com/JuliaPlots/RecipesBase.jl/blob/master/src/RecipesBase.jl#L245) macro in the RecipesBase package, even though we generally try to change the code there as little as possible. Given that normally one calls
```julia
@recipe function f(x::T)
    # recipe implementation
end
```
I imagine the recipe macro could call a `convert_arguments(T, x.args...)`, for example:
```julia
# this should be added by recipe macro
# and converted_args defined in RecipesBase
x.converted_args = convert_arguments(T, x.args)
```
so that one can use `x.converted_args` directly in the recipe implementation. But this is a big design change, so I'm not sure whether we should go for it.

mkborregaard:
couldn't this simply be resolved by dispatching the type recipe to the `recipetype(:corrplot)` syntax?

asinghvi17:
Not if you want the type to return a different thing when plotting histogram, scatter, or corrplot.

mkborregaard:
why not, sorry?

asinghvi17:
I may want plot(mytype) to dispatch to histogram instead, but still allow it to be compatible with corrplot




Sincerely,




Anshul Singhvi.







On Mon, Aug 19, 2019 at 7:09 AM -0400, "Michael Krabbe Borregaard" <notifications@github.com> wrote:












why not, sorry?

mkborregaard:
I don't understand completely why this cannot be achieved within the current framewor. The PR here looks harmless enough, though.

asinghvi17:
Let's suppose that I have a type `MyType`, which can be plotted as a histogram, a correlation matrix plot, and some other series types as well.

Now, if I define a type recipe for `MyType`, which dispatches to corrplot, then if I say `plot(MyType())`, I'm locked in to using corrplot - there's no way that I can define a conversion from my type to a corrplot matrix, _using the recipe system_ (for good kwarg defaults and all that), when the user invokes `corrplot(MyType())`, but have it converted to an array of `x`s (for histogram) when `histogram(MyType())` is called.  Perhaps `histogram` is a bad example, since it's a seriestype and can be detected, but `marginalhist` would not work in this way.

Even more importantly, if within a recipe I have an `@series` block like so (within a series recipe):
```
@series begin
    seriestype := histogram
    (MyType())
end
```
as far as I know, there is no way for that to be converted into a correct histogram.

asinghvi17:
@mkborregaard, I'll merge this for now, since you seem to be amenable to these changes, but this is a discussion we should have.  I've opened an RFC at https://github.com/JuliaPlots/Plots.jl/issues/2144, and I'll paste some of these comments there to flesh it out.

PR #269, opened 2019-11-18
Added silhouette plot

tobydriscoll:
This goes to address issue #268.


src/silhouetteplot.jl
```diff
@@ -0,0 +1,83 @@
...
+	a = assignments(R)  # assignments to clusters
+	c = counts(R)  # cluster sizes
+	k = length(c)  # number of clusters
+	n = sum(c)     # number of points overall
+
+	s = silhouettes(R,D)
+
+	# Settings for the axes
+	legend --> false
+	yflip := true
```
mkborregaard: Why do you override settings for these three attributes? Might `-->` not be better?

src/silhouetteplot.jl
```diff
@@ -0,0 +1,83 @@
...
+	xlims := [min(-0.1,minimum(s)),1]
+	# y ticks used to show cluster boundaries, and labels to show the sizes
+	yticks := cumsum([0;c]),["0",["+$z" for z in c]...]
+
+	# Generate the polygons for each cluster.
+	offset = 0;
+	plt = plot([],label="")
+	for i in 1:k
+		idx = (a.==i)    # members of cluster i
+		si = sort(s[idx],rev=true)
```
mkborregaard: better to use `filter` here which avoids allocating the temporary `idx` array

src/silhouetteplot.jl
```diff
@@ -0,0 +1,83 @@
...
+		@series begin
+			linealpha --> 0
+			seriestype := :shape
```
mkborregaard: why do you use `shape`? I think just a line with a `fillto` argument might be simpler (try to run `plot(sort(randn(1000)), fillto = 0)` to see the effect) and easier?

src/silhouetteplot.jl
```diff
@@ -0,0 +1,83 @@
...
+			label := "$i"
+			x = [0;repeat(si,inner=(2));0]
+			y = offset .+ repeat(0:c[i],inner=(2))
+			x,y
+		end
+		# text label to the left of the bars
+		@series begin
+			linealpha := 0
```
mkborregaard: or `seriesalpha?`. Consider `primary := false`

src/silhouetteplot.jl
```diff
@@ -0,0 +1,83 @@
...
+		end
+		offset += c[i];
+	end
+
+	# Dashed line for overall average.
+	savg = sum(s)/n
+	@series begin
+		linecolor := :black
+		linestyle := :dash
+		label := ""
```
mkborregaard: better `primary := false`

src/silhouetteplot.jl
tobydriscoll: `yflip` needs to be known in order to get the sorting order right. I'm not sure how to read that attribute if it wasn't set explicitly, and I thought top-to-bottom was most natural.

`xlims` and `yticks` should probably be "soft" assignments, I agree. Will modify.

src/silhouetteplot.jl
tobydriscoll: Sorry, I don't know how that works when the filtered array is not the one on which the selection is predicated. It's a boolean vector, so it's probably not creating a big performance issue.

src/silhouetteplot.jl
tobydriscoll: Does `fillto` work horizontally?

src/silhouetteplot.jl
tobydriscoll: OK.

src/silhouetteplot.jl
tobydriscoll: OK.

src/silhouetteplot.jl
tobydriscoll: Actually, I've learned that if I don't assign `yticks:=`, the labels don't get preserved. I'd really rather show the size of each cluster than the cumulative sizes. Do you know a fix?

asinghvi17: You can get the value by `plotattributes[:yticks]`, I think.  You can also read it, and then explicitly assign what you read to it.

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.12.0"
 Clustering = "aaaa29a8-35af-508c-8bc3-b662a17a0fe5"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 DataValues = "e7dc6d0d-1eca-5fa6-8ad6-5aecde8b7ea5"
+Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
```
asinghvi17: Needs compat bounds - I think "0.8" should be sufficient until the next release, where we could revisit the bound?

src/silhouetteplot.jl
tobydriscoll: No, if I try to reference `plotattributes[:yticks]` when it was left unset in the invoking plot command, I get a "key not found" error. I suppose that unset attributes aren't given values until later in the pipeline.

Project.toml
tobydriscoll: I have no idea, really, so that seems safe enough to me. I'm not using exotic functionality, it would seem to me.

src/silhouetteplot.jl
tobydriscoll: I've just avoided the issue by making `yflip` a keyword setting in `silhouetteplot`. It's true by default, but setting it to false is respected.

tobydriscoll:
Can I get an update on the changes requested to this PR?

mkborregaard:
@tobydriscoll really sorry for the ridiculously late response here. Could you rebase this onto master resolving the conflicts, and add an example to the Readme documenting the use of this function? Then I'll merge. Thanks, and sorry again.

tobydriscoll:
Updated the example in documentation string and the README.

PR #271, opened 2019-11-22
Allow for horizontal dendrogram 

oheil:
This allows for horizontal dendrograms, which is completely rotated 90 degree clockwise.

There is another (in my opinion not complete) solution to this in this PR:
https://github.com/JuliaPlots/StatsPlots.jl/pull/214
It is missing adjustments to the axes which is needed in some circumstances.
On the other hand, above PR is aimed for a better user experience with function clustermap.

I also added a complete example to the README.md for others to be able to create the heatmap with cluster dendrograms.
I had some problems to integrate the example image to the README.md so maybe this makes some effort on your side if you decide to merge this PR. Sorry but I didn't found the proper way to do this.

See also discussion
https://discourse.julialang.org/t/statsplots-dendrogram-on-the-right-side-or-rotated/31348


oheil:
Something missing here?

src/dendrogram.jl
```diff
@@ -1,4 +1,5 @@
-function treepositions(hc::Hclust, useheight::Bool)
+function treepositions(hc::Hclust, useheight::Bool, orientation=:v)
```
BeastyBlacksmith, line 1: In Plots.jl we use the symbols `:vertical` and `:horizontal` for `orientation

src/dendrogram.jl
```diff
@@ -17,22 +18,39 @@ function treepositions(hc::Hclust, useheight::Bool)
...
-    return xs, ys
+    if orientation == :h
+        return ys, xs
+    else
+        return xs, ys
+    end
 end
 
-@recipe function f(hc::Hclust; useheight=true)
+@recipe function f(hc::Hclust; useheight=true, orientation=:v )
```
BeastyBlacksmith, line 28: how does this interact with the `orientation` keyword that already exists?

src/dendrogram.jl
```diff
@@ -17,22 +18,39 @@ function treepositions(hc::Hclust, useheight::Bool)
...
-    treepositions(hc, useheight)
+    if orientation == :h
+        if useheight
+            hs=sum(hc.heights)
+            xlims --> (0, hs+hs*0.01)
+        else
+            xlims --> (0, Inf)
+        end
+        xshowaxis --> useheight
+        yforeground_color_axis := :white
```
BeastyBlacksmith, line 42: IMO all `:=`s should be `-->`s in this recipe or is there an apparent reason that this should be non-overridable?

oheil: Done as you said. Except for legend, as it was in the original code and it seems to make sense to disallow legends here.

src/dendrogram.jl
oheil: This I can't answer. I hope there is no side effect.

src/dendrogram.jl
oheil: Changed. Don't know anymore why I used :v and :h

oheil:
conflict is no conflict anymore.

src/dendrogram.jl
BeastyBlacksmith: I'm fine with providing no legend by default, but I think if a user wants to have a legend, there should be a way to get one

oheil: Of course. I have no opinion here, it is what the original author decided. Anyways, I changed it.

src/dendrogram.jl
BeastyBlacksmith: Does this flip the parabola or not?
```
plot([hc_row, (1:5).^2],yticks=false,xrotation=90,orientation=:horizontal)
```

oheil: No, it's just not working:
```
julia> plot([hc_row, (1:5).^2],yticks=false,xrotation=90,orientation=:horizontal)
ERROR: Cannot convert Hclust{Float64} to series data for plotting
```
This behaviour isn't changed. This doesn't work in current StatsPlots neither.

BeastyBlacksmith: hmm.. can't think of a good testcase here, I guess its fine. @dashw, you got objections?

oheil: I think it's is very specific for dendrograms and should not radiate outsides.

daschw:
Looks good to me, but there still seems to be an unresolved conflict.

oheil:
The current master has
`xticks --> (1:nnodes(hc), string.(1:nnodes(hc))[hc.order])`
I changed the new version to reflect this, but the conflict remains.
Maybe it must be resolved manually?

oheil:
The conflict lines shown are now in the `else `part. For the `if `part the lines changed to the y variants. 
So I would say, no conflict, but intended.

daschw:
I resolved the conflict now. @oheil could you have a look if the current version has the correct changes?

oheil:
Do you mean "master branch on oheil/StatsPlots.jl is correct ?" ? Yes, it is correct.

oheil:
The file diffs look correct too.

daschw:
Thanks, and thanks for your contribution!

PR #294, opened 2020-02-03
Add ecdfplot

sethaxen:
This PR adds recipes for plotting Empirical CDFs. Specifically, it adds a recipe for plotting `StatsBase.ECDF` and an `ecdfplot` plotting function.

Here's an example:

```julia
using StatsPlots, StatsBase, Distributions, Random
Random.seed!(42)

plot(ecdf(randn(100)), label = "Normal")
ecdfplot!(rand(Cauchy(), 100); label = "Cauchy")
plot!(ylabel = "Cumulative Probability", title = "Empirical CDF")
```

ecdf


src/ecdf.jl
```diff
@@ -0,0 +1,21 @@
+
+# ---------------------------------------------------------------------------
+# empirical CDF
+
+@recipe function f(ecdf::StatsBase.ECDF)
+    seriestype --> :path
+    linetype := :steppost
```
mkborregaard: These two seem to contradict each other - does `linetype := :steppost` have an effect?

sethaxen: Oh, I hadn't seen that `linetype` was an alias for `seriestype`. I removed the first definition, as `linetype := :steppost` is required for the plot to jump in the right places.

mkborregaard:
Great, thanks

PR #302, opened 2020-02-15
Fix https://github.com/JuliaPlots/StatsPlots.jl/issues/301

devmotion:
With this PR, the MWE in https://github.com/JuliaPlots/StatsPlots.jl/issues/301 can be run successfully. An example plot is
test

Moreover, I extracted `add_sym!` since it does not seem necessary to define these closures (and it fixes a tiny type instability).


daschw:
Great!

src/df.jl
```diff
@@ -188,17 +200,9 @@ function extract_columns_and_names(df, syms...)
...
-    add_sym!(s::Symbol) = s in names && push!(selected_cols, s)
-    # recursively extract column names
-    add_sym!(s) = foreach(add_sym!, s)
-
-    foreach(add_sym!, syms)
+    # extract selected column names
+    selected_cols = add_sym!(Symbol[], syms, names)
 
-    cols = columntable(select(df, unique(selected_cols)...))
+    cols = (; (s => getcolumn(df, s) for s in unique(selected_cols))...)
```
piever: This assumes that `df` implements `get_column`, which is not necessarily true (it could be a row-based table with no column access). It's probably easier to stick to the old code but using `TableOperations.select` (`select` moved to https://github.com/JuliaData/TableOperations.jl).

daschw:
Thanks a lot @devmotion and @piever !

Issue #140, opened 2018-04-13
How to plot legend upside down in stacked groupedbar


ngphuoc:
It's is easier to see the legend and the stacked bars colors having the same order. Especially when there are many groups.

mkborregaard:
Yes, they should be the same color for sure. You can't specify it, but it should be built in.

Lilyhanig:
The problem is that the order is reverse to the way it get's stacked, they're listed top down, when generally bottom up is preferred for stacked bars.

mkborregaard:
Yes exactly. The order of colors in the bars should be reversed. It's a really nice first PR if anyone is interested.

mkborregaard:
Closed in #309

PR #309, opened 2020-02-29
Fix groupedbar ordering

Moelf:
Address #140 
```
groupedbar([[3,3] [2,-2] [1,0.5]], bar_position=:stack)
```
Before:

After:

### DIFF
--- a/src/bar.jl
+++ b/src/bar.jl
@@ -59,16 +59,20 @@ end
 
 function groupedbar_fillrange(y)
     nr, nc = size(y)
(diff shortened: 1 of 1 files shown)


mkborregaard:
Nice change, thanks

MaximilianJHuber:
This changed the ordering even for plots that have no negative bars. On the [GitHub page](https://github.com/JuliaPlots/StatsPlots.jl#grouped-bar-plots) it is:
image
With StatsPlots 0.14.6:
image

PR #315, opened 2020-03-07
Add Stacked Histogram

Moelf:
```
stackedhist([randn(5000), randn(1000)], bins=-3:0.2:3)
tt = fit(Histogram, rand(2000), 0:0.2:1);  tt2 = fit(Histogram, rand(4000), 0:0.2:1)
stackedhist([tt, tt2])
```

stackhist1
stackhist2

Suggestion welcomed.

### DIFF
--- a/src/stackedhist.jl
+++ b/src/stackedhist.jl
@@ -1,0 +1,32 @@
+@userplot StackedHist
+
+@recipe function f(h::StackedHist)
+    arrays = reverse(h.args[1])
+    elt = eltype(arrays)
+    if elt <: Histogram 
+        bins = arrays[1].edges[1] # edges is a tuple
+        weights_ary = hcat([x.weights for x in arrays]...) #reverse so it stacks from top to bottom
+    elseif elt <: Array{<:Real} # if raw datas
+        # if no bins given, guess by fitting first data array
+        bins = pop!(plotattributes, :bins, fit(Histogram, arrays[1]).edges[1])
+        hist_wgts = pop!(plotattributes, :weights, [nothing]) |> reverse
+        weights_ary = (
+            hist_wgts == [nothing] ?
+                [fit(Histogram, x, bins).weights for x in arrays] :
+                [fit(Histogram, x, weights(y), bins).weights for (x, y) in zip(arrays, hist_wgts)]
+            )
+        weights_ary = hcat(weights_ary...)
+            
+    else
+        error("args[1] should be collection of data arrays or of histograms")
+    end
+    
+    stacked_data = abs.(reverse(cumsum(weights_ary, dims = 2), dims=2))
+    ymin = 10^floor(log10(minimum(stacked_data)))
+    stacked_data = vcat(stacked_data, zeros(1, size(stacked_data)[2])) # FIXME janky make the last bin visible
+        
+    fillrange --> 10^-6
+    ylim --> (ymin>=1 ? (10^0, Inf) : (ymin/10, Inf))
+    seriestype := :steppost
+    bins, stacked_data
+end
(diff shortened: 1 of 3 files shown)


src/stackedhist.jl
```diff
@@ -0,0 +1,32 @@
...
+
+@recipe function f(h::StackedHist)
+    arrays = reverse(h.args[1])
+    elt = eltype(arrays)
+    if elt <: Histogram 
+        bins = arrays[1].edges[1] # edges is a tuple
+        weights_ary = hcat([x.weights for x in arrays]...) #reverse so it stacks from top to bottom
+    elseif elt <: Array{<:Real} # if raw datas
+        # if no bins given, guess by fitting first data array
+        bins = pop!(plotattributes, :bins, fit(Histogram, arrays[1]).edges[1])
```
mkborregaard: I'd recommend using the histogram binning mechanism in Plots instead https://github.com/JuliaPlots/Plots.jl/blob/master/src/recipes.jl#L462-L834 it's much better and has taken considerable effort. And `bins` can also be a symbol referring to alternative binning methods. Well actually the best might be to to take the binning functionality from Plots (and there's also some here in StatsPlots) and PR it upstream to StatsBase.

src/stackedhist.jl
```diff
@@ -0,0 +1,32 @@
...
+    else
+        error("args[1] should be collection of data arrays or of histograms")
+    end
+    
+    stacked_data = abs.(reverse(cumsum(weights_ary, dims = 2), dims=2))
+    ymin = 10^floor(log10(minimum(stacked_data)))
+    stacked_data = vcat(stacked_data, zeros(1, size(stacked_data)[2])) # FIXME janky make the last bin visible
+        
+    fillrange --> 10^-6
+    ylim --> (ymin>=1 ? (10^0, Inf) : (ymin/10, Inf))
```
mkborregaard: why is this needed?

src/stackedhist.jl
```diff
@@ -0,0 +1,32 @@
...
+                [fit(Histogram, x, weights(y), bins).weights for (x, y) in zip(arrays, hist_wgts)]
+            )
+        weights_ary = hcat(weights_ary...)
+            
+    else
+        error("args[1] should be collection of data arrays or of histograms")
+    end
+    
+    stacked_data = abs.(reverse(cumsum(weights_ary, dims = 2), dims=2))
+    ymin = 10^floor(log10(minimum(stacked_data)))
```
mkborregaard: sorry not quite getting the idea with the exponents of 10?

mkborregaard:
Looks nice - I've got a few comments above.

src/stackedhist.jl
Moelf: 0 is at `-Inf` in log scale

src/stackedhist.jl
Moelf: `ylim` is to avoid it goes wrong in logscale:
`ArgumentError: At least one finite value must be provided to formatter.`

Moelf:
at the moment there's visual glitch when content is very small && in log scale when format is `svg`
`stackedhist([abs.(rand(100)), abs.(rand(20))], fill=true, yscale=:log10)`
image
`stackedhist([abs.(rand(100)), abs.(rand(20))], fill=true, yscale=:log10, fmt=:png)`
image

src/stackedhist.jl
Moelf: ah, I think it all comes down to
https://github.com/JuliaPlots/Plots.jl/blob/bece5a2620700c2d2d4955732a600c8222674c50/src/recipes.jl#L702

mkborregaard:
@Moelf sorry but it looks like this has been superseded by #319?

PR #329, opened 2020-04-14
More macro hygiene in `@df`

asinghvi17:
I have a package which imports but does not export StatsPlots; in this case, the `@df` macro fails.  A quick fix in this instance is to `export StatsPlots`, but if this module is located within another, that becomes untenable.

This PR simply changes all references to `StatsPlots` in generated code to `$(@__MODULE__)`.  This means that the "true location" of StatsPlots, which may be `Module1.ModuleZ.HelloWorld.StatsPlots`, is used, instead of assuming that `StatsPlots` is available in the calling module.


asinghvi17:
Mind if I make a minor release after merging this?  I think we also need to get a Clustering compat bump out.

PR #344, opened 2020-06-09
Covariance ellipse recipe

yha:
A recipe for plotting a 2×2 covariance matrix as an ellipse:
```
covellipse([0,2], [2 1; 1 4], n_std=2, aspect_ratio=1, label="cov1")
covellipse!([1,0], [1 -0.5; -0.5 3], showaxes=true, label="cov2")
```
tmp


mkborregaard:
Nice. What about correlation matrix?

yha:
How do you suppose to visualize a correlation matrix? Compute the variance matrix from correlations and marginal variances and plot that? Something else?

mkborregaard:
I was thinking of something like this, from R's corrplot package

Skærmbillede 2020-06-10 kl  08 19 23

mkborregaard:
Might be unrelated, just thinking out loud

yha:
I think it shouldn't be too hard to write something like this on top of the `covellipse` recipe.

PR #359, opened 2020-07-30
fix df macro

BeastyBlacksmith:
Since plots does not convert Symbols to Strings anymore the macro needs to do this.

Do we need to change compat here?

Fix https://github.com/JuliaPlots/StatsPlots.jl/issues/358

### DIFF
--- a/src/df.jl
+++ b/src/df.jl
@@ -152,7 +152,7 @@ function add_label(argnames, f, args...; kwargs...)
         if (i === nothing)
             return f(args...; kwargs...)
         else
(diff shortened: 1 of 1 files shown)


daschw:
> Since plots does not convert Symbols to Strings anymore the macro needs to do this.

Do you know when this change happened in Plots?

BeastyBlacksmith:
That was https://github.com/JuliaPlots/Plots.jl/pull/2837, so from v1.5.2 onwards

daschw:
But this change should also work for the Plots versions before, right? I don't think we need a compat change here.

Issue #357, opened 2020-07-29
mismatch between data an bar height in grouped bar chart (log+lin, only in master)


joergbuchwald:
Hi,
since an update recently, I discovered a serious mismatch between the bar height and the data producing the bars.
Here are some examples:
https://github.com/joergbuchwald/projects/blob/master/temp/logplot.ipynb
In the log plot, all bars do have nearly the same height whereas there should be one order of magnitude in between.
The same holds also for the linear and log plot below. The ticks on the left should match the bar heights (e.g. for 10 and 12)

joergbuchwald:
Updating to julia 1.5 seems to fix the problem.
However, #263 seems to be somehow relevant again: using `log10` instead of `log` requires the `fillrange` attribute to be set again.

hendri54:
I don't think it's just in master. With Julia 1.4.2 (MacOS), Plots 1.5.8, StatsPlots 0.14.7, this simple example gives bars that are taller than 1:

```
julia> using Random, StatsPlots
julia> rng = MersenneTwister(123);
julia> p=groupedbar(["1", "2", "3", "4"], rand(rng, 4, 2))
```

joergbuchwald:
Yes, probably. I haven't seen this.
It worked with 0.14.6, as I remember. The 0.14.7 release is quite new.

hendri54:
Thanks for that pointer - pinning 0.14.6 works for me.

BeastyBlacksmith:
Would be nice, if you can report back, if 0.14.8 solves this issue, when it is released

joergbuchwald:
It doesn't seem to be fixed yet.
(See the above mentioned jupyter notebook for further details).

joergbuchwald:
This looks good! Thank you very much!

Issue #355, opened 2020-07-14
groupedbar() undesired stacking


yookiwooki:
I'm working on the master branch, and I'm finding undesired stacking.  I'm seeing this output with the groupedbar() example from the Readme.  Since I'm using rand() there shouldn't be any values above 1.  Anybody else seeing this?

```julia
ctg = repeat(["Category 1", "Category 2"], inner = 5)
nam = repeat("G" .* string.(1:5), outer = 2)

groupedbar(nam, rand(5, 2), group = ctg, xlabel = "Groups", ylabel = "Scores",
        title = "Scores by group and category", bar_width = 0.67,
        lw = 0, framestyle = :box)
```
<img width="401" alt="Capture" src="https://user-images.githubusercontent.com/33902914/87463408-f1e1c600-c5d6-11ea-9dba-bf9e71f26855.PNG">

SteffenPL:
This might be related... In my example with `groupedhist`.

I don't even see the relationship between the combined plot and the single ones...

```julia
using DataFrames, Random, StatsPlots, Query
Random.seed!(2)
df = DataFrame(a = (ceil.(5 * rand(1000))), g=(rand(1000) .> 0.5))
p1 = df |> @filter(_.g == false) |> @df groupedhist(:a, group=:g, color=1)
p2 = df |> @filter(_.g == true) |> @df groupedhist(:a, group=:g, color=2)
p12 = @df df groupedhist(:a, group=:g, nbins=5)
plot(p1, p12, p2, layout=(1,3), link=:both, legend=:bottomright)
```

![groupedhist](https://user-images.githubusercontent.com/4634605/88113435-167b0680-cbb2-11ea-8207-903280fc54af.png)

SteffenPL:
See also #219

SteffenPL:
An easier example, somehow the groups get mixed up.

```julia
using DataFrames, Random, StatsPlots, Query
df = DataFrame(a = [1,1,1,1,2,1], g= (collect(1:6) .< 5))

6×2 DataFrame
│ Row │ a     │ g    │
│     │ Int64 │ Bool │
├─────┼───────┼──────┤
│ 1   │ 1     │ 1    │
│ 2   │ 1     │ 1    │
│ 3   │ 1     │ 1    │
│ 4   │ 1     │ 1    │
│ 5   │ 2     │ 0    │
│ 6   │ 1     │ 0    │


[...]

PR #364, opened 2020-08-07
Roll back to 0.14.6

BeastyBlacksmith:
This is a full rollback to 0.14.6 with the df macro fix appended.

We need some tests for #357 and #355 that don't rely on the RNG to figure out which commit caused this.
Tagging all the authors of the involved commits:
@Moelf, @diegozea, @PallHaraldsson, @xukai92, @yha, @HenriDeh


BeastyBlacksmith:
Well the rollback didn't really work out ^^;

PR #368, opened 2020-08-10
Fix upper limit of groupedbar

BeastyBlacksmith:
Fix https://github.com/JuliaPlots/StatsPlots.jl/issues/357

### DIFF
--- a/src/bar.jl
+++ b/src/bar.jl
@@ -48,15 +48,9 @@ grouped_xy(y::AbstractArray) = 1:size(y,1), y
     end
 
     # compute fillrange
-    y, fr = isstack ? groupedbar_fillrange(y) : (y, get(plotattributes, :fillrange, nothing))
+    # y, fr = groupedbar_fillrange(y)
     fillrange := if isylog
         fill_bottom = min(minimum(y) / 100, the_ylims[1])
-        if isstack
-            # to fill the blank of the bottom segment depeding on if ylims
-            replace!(fr, 0 => fill_bottom)
-        else
-            fill_bottom
-        end
     else
         get(plotattributes, :fillrange, nothing)
     end
--- a/src/bar.jl
+++ b/src/bar.jl
@@ -47,26 +47,34 @@ grouped_xy(y::AbstractArray) = 1:size(y,1), y
         xmat
     end
 
-    # compute fillrange
-    # y, fr = groupedbar_fillrange(y)
-    fillrange := if isylog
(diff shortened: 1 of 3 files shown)


src/bar.jl
```diff
@@ -48,15 +48,9 @@ grouped_xy(y::AbstractArray) = 1:size(y,1), y
     end
 
     # compute fillrange
-    y, fr = isstack ? groupedbar_fillrange(y) : (y, get(plotattributes, :fillrange, nothing))
```
daschw, line 51: So how are the y positions and fillranges for stacked bars calculated now?

BeastyBlacksmith: It's right below:
```julia
fill_bottom = min(minimum(y) / 100, the_ylims[1])
```

BeastyBlacksmith: Happy to have more test cases though

daschw: But they are different for `:stack`ed groupedbars, which is calculated in `groupedbar_fillrange`. These changes only cover the `:dodge` case.

daschw:
```julia
using StatsPlots

x = [
    1 2 2
    3 2 1
]

groupedbar(x, bar_position = :stack)
```
With this PR:
stackedbar

On master:
stackedbar_master

BeastyBlacksmith: should be fixed now

src/bar.jl
```diff
@@ -47,32 +47,34 @@ grouped_xy(y::AbstractArray) = 1:size(y,1), y
...
-function groupedbar_fillrange(y)
+function groupedbar_fillrange(y; y_bottom = 0)
     nr, nc = size(y)
     # bar series fills from y[nr, nc] to fr[nr, nc], y .>= fr
     fr = zeros(nr, nc)
     y = copy(y)
     y[.!isfinite.(y)] .= 0
     for r = 1:nr
-        y_neg = 0
+        y_neg = y_bottom
```
daschw, line 77: I don't think this makes sense. This only offsets all negative values by `+y_bottom`. Can't we leave `groupedbar_fillrange` as it was? Isn't this covered by L62 already? Or, respectively, would we even support negative input for log scales?

BeastyBlacksmith: you are right, thats a leftover, I reverted this in https://github.com/JuliaPlots/StatsPlots.jl/commit/bcc35353482e596e32808140ca1257b56164b55c

PR #388, opened 2020-10-21
Adding a marginalkde plot much like marginalhist

farr:
This PR adds a `marginalkde` plot much like the `marginalhist` plot.


BeastyBlacksmith:
Can you show an example?

farr:
```julia
marginalkde(randn(1024), randn(1024))
```

marginalkde

farr:
Or, maybe more illustrative:

```julia
x = randn(1024)
y = randn(1024)
marginalkde(x, x+y)
```

correlated-marg

src/marginalkde.jl
```diff
@@ -0,0 +1,73 @@
...
+    x,y = kc.args
+
+    x = vec(x)
+    y = vec(y)
+
+    m_x = median(x)
+    m_y = median(y)
+
+    dx_l = m_x - quantile(x, 0.16)
+    dx_h = quantile(x, 0.84) - m_x
```
daschw, line 13: Are 0.16 and 0.84 the best choice in all cases, or should we add an additional kwarg or alternatively specify `xlims` and `ylims` using `-->` in the recipe?

farr: Hmm.  That's controllable via the `clip` argument (which multiplies these values); those numbers would represent +/- 1-sigma in a Gaussian, so are probably sensible for choosing a "scale" that gets multiplied by the `clip` argument.

daschw: Oh, sorry, I did not take a close enough look.

daschw (approved):
Looks great!

farr:
Great---thanks!

PR #394, opened 2020-11-09
Fix boxplot labels

### LINKED ISSUE #392: Boxplot and violinplot example and boxplot markers
On Julia 1.5.2 StatsPlots v0.14.15, I run the `boxplot, dotplot, and violin` example code
```
using StatsPlots, DataFrames, RDatasets

singers = dataset("lattice","singer")
@df singers violin(:VoicePart,:Height,marker=(0.2,:blue,stroke(0)))
@df singers boxplot!(:VoicePart,:Height,marker=(0.3,:orange,stroke(2)), alpha=0.75)
@df singers dotplot!(:VoicePart,:Height,marker=(:black,stroke(0)))
```
and get the following error:
```
Cannot convert CategoricalValue{String,UInt8} to series data for plotting
```

Using `convert(Array{String}:VoicePart)` instead of `:VoicePart` seems to fix that problem.

**However**, the resulting boxplots have markers at the end of each path segment, which is wrong:

image

This seems to be "fixed" by removing the `marker` argument: 
```@df singers boxplot!(convert(Array{String},:VoicePart),:Height,alpha=0.75)```
image

Could the following code maybe play a role? 
https://github.com/JuliaPlots/StatsPlots.jl/blob/7927de64edd7daf7418e4f457372aa931b05cba2/src/boxplot.jl#L174

### PR DESCRIPTION
**Pull request by daschw**
```julia
x = repeat(1:4, inner=30)
y = randn(120)
g = repeat(1:3, outer=40)
groupedboxplot(x, y, group=g)
```
before
groupedboxplot_before
now
groupedboxplot_now

EDIT: close #392

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StatsPlots"
 uuid = "f3b207a7-027a-5e70-b257-86293d7955fd"
-version = "0.14.15"
+version = "0.14.16"
 
 [deps]
 Clustering = "aaaa29a8-35af-508c-8bc3-b662a17a0fe5"
--- a/README.md
+++ b/README.md
@@ -201,23 +201,23 @@ cornerplot(M, compact=true)
 
 ```julia
 import RDatasets
-singers = RDatasets.dataset("lattice","singer")
-@df singers violin(:VoicePart,:Height,marker=(0.2,:blue,stroke(0)))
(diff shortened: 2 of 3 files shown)


daschw:
updated readme images:
readme_singers
readme_singers_2

PR #422, opened 2021-02-07
Use cumsum in the cdensity recipe

sethaxen:
The recipe for `cdensity` computes `cumsum` but allocating all subarrays. This is incredibly wasteful. This PR uses `cumsum`. For a vector of length 1000, this part of the recipe is 100x faster.


daschw:
Nice, thanks!

PR #434, opened 2021-03-30
Change away from `scatterpath` to minimise confusion

nilshg:
`scatterpath` leads to lines being a different colour from markers, and the markers can end up being the same colour as different lines on the same plot, making the legend very confusing.

### DIFF
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -27,7 +27,7 @@ yz_args(dist::Distribution{N, T}) where N where T<:Discrete = (UnitRange(Int.(de
 # this "user recipe" adds a default x vector based on the distribution's μ and σ
 @recipe function f(dist::Distribution)
     if dist isa Distribution{Univariate,Discrete}
-        seriestype --> :scatterpath
+        marker --> :square
     end
     (dist, yz_args(dist)...)
 end
@@ -49,7 +49,7 @@ end
         @series begin
             seriesargs = isempty(yz) ? yz_args(di) : yz
             if di isa Distribution{Univariate,Discrete}
-                seriestype --> :scatterpath
+                marker --> :square
             end
             (di, seriesargs...)
         end


nilshg:
I've initially raised this on Zulip but thought I'd just PR straight away - this essentially moves us from
image

to 
image

BeastyBlacksmith:
If we are to go with this PR I would leave the marker be `:circle`, but actually I would prefer to fix scatterpath to have the same color for markers and line

nilshg:
That's a very good point, I generally can't see why one would want different colours when plotting a `scatterpath`.

Is it this recipe here: https://github.com/JuliaPlots/Plots.jl/blob/0a84ff8caff94e1923d379465d14998835e24d18/src/recipes.jl#L164? 

I just monkey-patched it locally by adding `color := 1` and that seems to do the trick, but I'm assuming that's a bit heavy handed. Not sure how to tell it to use the same color for both series without fixing the color?

BeastyBlacksmith:
`color = plotattr[:seriescolor]` might work

nilshg:
Sorry, can't get that to work - `color := plotattr[:seriescolor]` complains that there's `no method matching getindex(::typeof(plotattr), ::Symbol)` (is `plotattr` a function?), and any other changes I make to try and align it with what's in here give me a pretty inscrutable `ERROR: The seriestype didn't change in series recipe scatterpath. This will cause a StackOverflow.` Looks like this is beyond me!

BeastyBlacksmith:
should be a dictionary, but maybe the name was `plotattributes`? Its some time I worked on this ^^;

BeastyBlacksmith:
may be fixed by https://github.com/JuliaPlots/Plots.jl/commit/64f3d589b68cb1e227f28940375045ee3dfaf481

PR #440, opened 2021-04-21
Updates boxplot to allow for custom percentiles

MattAmos:
Adds the ability for custom percentiles to be passed to boxplots (and grouped box plots). Retains current default behaviour of using quartiles when custom percentiles are omitted.

### DIFF
--- a/src/boxplot.jl
+++ b/src/boxplot.jl
@@ -12,7 +12,8 @@ notch_width(q2, q4, N) = 1.58 * (q4-q2)/sqrt(N)
     notch = false,
     range = 1.5,
     outliers = true,
-    whisker_width = :half
+    whisker_width = :half,
+    percentiles = [0.25, 0.5, 0.75]
 )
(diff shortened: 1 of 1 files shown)


MattAmos:
Any feedback on this would be greatly appreciated.

BeastyBlacksmith:
Can you add some tests? Otherwise looks good to me

cristobaltapia:
@MattAmos thanks for this PR. I was wondering if it would be possible to also consider the placement of the whiskers as percentiles. As I understand this PR, it only allows to modify the extent of the box and add additional lines within it at the specified positions. Could we add an option like `whiskerpercentile=[0.05, 0.95]` that modifies the extension of the whiskers (currently controlled exclusively by `IQR * range`)?

MattAmos:
Thanks @BeastyBlacksmith and @cristobaltapia. I'll work on adding tests and configurable whiskers.

Ph0non:
Any updates on this?

PR #447, opened 2021-06-07
Fix dendrogram

diegozea:
Hi!

This fixes a minor bug I introduced in https://github.com/JuliaPlots/StatsPlots.jl/pull/356 to match the Wikipedia example. I'm sorry for that. In that example, the merged nodes' height is the **branch length**, half of the distance between the merged nodes for the ultrametric tree. That's problematic in this context, where `heights[i]` is already the `height` at which the merge node `i' should be placed. Both *Clustering.jl* and *R* considered that the height is the actual **distance** rather than the brach length, which makes much sense in combination with `hline!` and `cutree` (this is how I've realised this bug). With this simple fix, our plot matches the R plot :) Here is the [wiki example](https://en.wikipedia.org/wiki/Complete-linkage_clustering):

image

This also fixes https://github.com/JuliaPlots/StatsPlots.jl/issues/224 

image

Many thanks for considering my PR,

Best regards,

P.S.: Maybe, in the future, we can have `useheight` to be `:distance` (the default, this PR), `:branch_length` (previous buggy behaviour that can be useful for phylogeny) or `:none` (the actual `false`). But this minor PR should fix the problem of not matching *R* and `cutree` for the moment.

### DIFF
--- a/src/dendrogram.jl
+++ b/src/dendrogram.jl
@@ -11,7 +11,7 @@ function treepositions(hc::Hclust, useheight::Bool, orientation=:vertical)
         x2, y2 = nodepos[hc.merges[i, 2]]
 
         xpos = (x1 + x2) / 2
-        ypos = useheight ?  (hc.heights[i] / 2) : (max(y1, y2) + 1)
+        ypos = useheight ?  hc.heights[i] : (max(y1, y2) + 1)
         
         nodepos[i] = (xpos, ypos)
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
It would be great to have some tests to keep this behaviour

diegozea:
Thanks for the suggestion, @BeastyBlacksmith; I have just added the test :)

PR #451, opened 2021-06-18
Improvements to plotting PMFs for discrete distributions

### LINKED ISSUE #287: Showing Probability Mass Function for discrete distributions
For [discrete distriubtions](https://juliastats.org/Distributions.jl/stable/univariate/#Discrete-Distributions-1), it would be nice to have a way to plot the [pmf](https://en.wikipedia.org/wiki/Probability_mass_function).

Currently, when trying to plot a discrete distribution, it shows a line plot with the two points connected. E.g.:

`plot(Bernoulli(0.1))`:
image

### LINKED ISSUE #450: Plotting discrete distributions with non-integer support
StatsPlots currently incorrectly assumes all discrete distributions have their support on the integers. This causes distributions with e.g. support on the floating point numbers to error when you plot them. A MWE:

```julia
julia> using StatsPlots, Distributions

julia> plot(Dirac(1.0))  # this is fine

julia> plot(Dirac(0.5))
ERROR: InexactError: Int64(0.5)
Stacktrace:
  [1] Int64
    @ ./float.jl:723 [inlined]
  [2] _broadcast_getindex_evalf
    @ ./broadcast.jl:648 [inlined]
  [3] _broadcast_getindex
    @ ./broadcast.jl:621 [inlined]
  [4] #19
    @ ./broadcast.jl:1098 [inlined]
  [5] ntuple
    @ ./ntuple.jl:49 [inlined]
  [6] copy
    @ ./broadcast.jl:1098 [inlined]
  [7] materialize
    @ ./broadcast.jl:883 [inlined]
  [8] yz_args(dist::Dirac{Float64})
    @ StatsPlots ~/.julia/packages/StatsPlots/ie0L8/src/distributions.jl:25
  [9] macro expansion
    @ ~/.julia/packages/StatsPlots/ie0L8/src/distributions.jl:32 [inlined]
 [10] apply_recipe(plotattributes::AbstractDict{Symbol, Any}, dist::Distribution)
    @ StatsPlots ~/.julia/packages/RecipesBase/92zOw/src/RecipesBase.jl:282
 [11] _process_userrecipes!(plt::Any, plotattributes::Any, args::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/CirY4/src/user_recipe.jl:36
 [12] recipe_pipeline!(plt::Any, plotattributes::Any, args::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/CirY4/src/RecipesPipeline.jl:70
 [13] _plot!(plt::Plots.Plot, plotattributes::Any, args::Any)
    @ Plots ~/.julia/packages/Plots/FKcum/src/plot.jl:172
 [14] plot(args::Any; kw::Any)
    @ Plots ~/.julia/packages/Plots/FKcum/src/plot.jl:58
 [15] plot(args::Any)
    @ Plots ~/.julia/packages/Plots/FKcum/src/plot.jl:52
 [16] top-level scope
    @ REPL[23]:1
```

This came up in https://github.com/devmotion/InferenceDiagnostics.jl/pull/7. @devmotion [noted](https://github.com/devmotion/InferenceDiagnostics.jl/pull/7#issuecomment-861038830):
> ...the resulting plot is incorrect since StatsPlots incorrectly assumes that discrete distributions are integer-valued (https://github.com/JuliaPlots/StatsPlots.jl/blob/6e65c64fb99ca5148989766092a5432391ab1e3b/src/distributions.jl#L25). This is not correct e.g. for discrete location-scale distributions, Dirac distributions, and `DiscreteNonParametric`.

sethaxen:
This PR fixes #450 and fixes #287.

### DIFF
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -25,17 +25,15 @@ end
 
 # this "user recipe" adds a default x vector based on the distribution's μ and σ
 @recipe function f(dist::Distribution)
-    if dist isa DiscreteUnivariateDistribution && get(plotattributes, :seriestype, :none) === :none
-        seriestype := :sticks
-        markershape --> :circle
+    if dist isa DiscreteUnivariateDistribution
+        seriestype --> :sticks
     end
     (dist, yz_args(dist)...)
 end
 
 @recipe function f(m::Distributions.MixtureModel; components = true)
-    if m isa DiscreteUnivariateDistribution && get(plotattributes, :seriestype, :none) === :none
-        seriestype := :sticks
-        markershape --> :circle
+    if m isa DiscreteUnivariateDistribution
+        seriestype --> :sticks
     end
     if components
         for c in m.components
@@ -52,9 +50,8 @@ end
     for di in distvec
         @series begin
             seriesargs = isempty(yz) ? yz_args(di) : yz
-            if di isa DiscreteUnivariateDistribution && get(plotattributes, :seriestype, :none) === :none
-                seriestype := :sticks
-                markershape --> :circle
+            if di isa DiscreteUnivariateDistribution
+                seriestype --> :sticks
             end
             (di, seriesargs...)
         end
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -35,7 +35,8 @@ end
 # this "user recipe" adds a default x vector based on the distribution's μ and σ
 @recipe function f(dist::Distribution)
     if dist isa Distribution{Univariate,Discrete}
-        seriestype --> :scatterpath
+        seriestype --> :sticks
+        markershape --> :circle
     end
     (dist, yz_args(dist)...)
 end
@@ -57,7 +58,8 @@ end
         @series begin
             seriesargs = isempty(yz) ? yz_args(di) : yz
             if di isa Distribution{Univariate,Discrete}
-                seriestype --> :scatterpath
+                seriestype --> :sticks
+                markershape --> :circle
             end
             (di, seriesargs...)
         end
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -42,6 +42,10 @@ end
 end
 
 @recipe function f(m::Distributions.MixtureModel; components = true)
+    if m isa Distribution{Univariate,Discrete} && get(plotattributes, :seriestype, :none) === :none
+        seriestype := :sticks
+        markershape --> :circle
+    end
     if components
         for c in m.components
             @series begin
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -34,8 +34,8 @@ end
 
 # this "user recipe" adds a default x vector based on the distribution's μ and σ
 @recipe function f(dist::Distribution)
-    if dist isa Distribution{Univariate,Discrete}
-        seriestype --> :sticks
+    if dist isa Distribution{Univariate,Discrete} && get(plotattributes, :seriestype, :none) === :none
+        seriestype := :sticks
         markershape --> :circle
     end
     (dist, yz_args(dist)...)
@@ -57,8 +57,8 @@ end
     for di in distvec
         @series begin
             seriesargs = isempty(yz) ? yz_args(di) : yz
-            if di isa Distribution{Univariate,Discrete}
-                seriestype --> :sticks
+            if di isa Distribution{Univariate,Discrete} && get(plotattributes, :seriestype, :none) === :none
+                seriestype := :sticks
                 markershape --> :circle
             end
             (di, seriesargs...)
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -3,7 +3,7 @@
 function default_range(dist::Distribution, alpha = 0.0001)
     minval = isfinite(minimum(dist)) ? minimum(dist) : quantile(dist, alpha)
     maxval = isfinite(maximum(dist)) ? maximum(dist) : quantile(dist, 1-alpha)
-    minval, maxval
+    promote(minval, maxval)
 end
 
 function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
@@ -18,7 +18,7 @@ function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
             maxval = thismax
         end
     end
-    minval, maxval
+    promote(minval, maxval)
 end
 
 yz_args(dist) = default_range(dist)
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -22,7 +22,23 @@ function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
 end
 
 yz_args(dist) = default_range(dist)
-yz_args(dist::Distribution{N, T}) where N where T<:Discrete = (UnitRange(Int.(default_range(dist))...),)
+function yz_args(dist::Distribution{N, T}) where N where T<:Discrete
+    xmin, xmax = default_range(dist)
+    try
+        sup = support(dist)
+        imin = findfirst(≥(xmin), sup)
+        imax = findlast(≤(xmax), sup)
+        sup_subrange = sup[imin:imax]
+        if sup_subrange isa AbstractVector
+            return (sup_subrange,)
+        else # handle e.g. when support is a tuple
+            return ([sup_subrange...],)
+        end
+    catch
+        # fallback when calling `support` fails
+        return (xmin:xmax,)
+    end
+end
 
 # this "user recipe" adds a default x vector based on the distribution's μ and σ
 @recipe function f(dist::Distribution)
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -23,20 +23,12 @@ end
 
 yz_args(dist) = default_range(dist)
 function yz_args(dist::Distribution{N, T}) where N where T<:Discrete
-    xmin, xmax = default_range(dist)
-    try
+    minval, maxval = extrema(dist)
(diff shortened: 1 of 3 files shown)


devmotion:
I think it would be easier (and probably faster) to not use `detault_range` but
- to look up `min_dist, max_dist = extrema(dist)` (that will always work)
- to use `support` if both are bounded (this should also work, the bug in Distributions only affects unbounded support)
- and to use the quantiles with a unit range in the other case (which is likely wrong in some cases but probably the best one can do right now)

src/distributions.jl
```diff
@@ -22,7 +22,15 @@ function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
 end
 
 yz_args(dist) = default_range(dist)
-yz_args(dist::Distribution{N, T}) where N where T<:Discrete = (UnitRange(Int.(default_range(dist))...),)
+function yz_args(dist::Distribution{N, T}) where N where T<:Discrete
```
devmotion, line 25:
```suggestion
function yz_args(dist::DiscreteUnivariateDistribution)
```

src/distributions.jl
```diff
@@ -22,7 +22,15 @@ function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
...
+    minval, maxval = extrema(dist)
+    if isfinite(minval) && isfinite(maxval)  # bounded
+        sup = support(dist)
+        return sup isa AbstractVector ? (sup,) : ([sup...],)
```
devmotion, line 29: I guess it doesn't matter since I assume it only handles `Dirac` but one could avoid splatting with
```suggestion
        return sup isa AbstractVector ? (sup,) : (reduce(vcat, sup),)
```

devmotion (commented):
I guess it's the best that can be done currently. Maybe it would be a good opportunity to address https://github.com/JuliaPlots/StatsPlots.jl/issues/287 as well.

sethaxen:
> Maybe it would be a good opportunity to address #287 as well.

Done!
```julia
julia> plot(Binomial(50, 0.5))
```
d

src/distributions.jl
```diff
@@ -22,12 +22,21 @@ function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
...
+        return (UnitRange(default_range(dist)...),)
+    end
+end
 
 # this "user recipe" adds a default x vector based on the distribution's μ and σ
 @recipe function f(dist::Distribution)
     if dist isa Distribution{Univariate,Discrete}
-        seriestype --> :scatterpath
+        seriestype --> :sticks
+        markershape --> :circle
```
sethaxen, line 39: Might be better to use some logic here to only set the `markershape` if `seriestype` hasn't been set. Otherwise the user asks for a `path` and we coerce it to show scatter points.

sethaxen:
I made the same work for discrete mixtures. e.g. here is zero-inflated Poisson:

```julia
julia> zip = MixtureModel([Dirac(0), Poisson(10)], [0.1, 0.9])
MixtureModel{Distribution{Univariate, Discrete}}(K = 2)
components[1] (prior = 0.1000): Dirac{Int64}(value=0)
components[2] (prior = 0.9000): Poisson{Float64}(λ=10.0)

julia> plot(zip)
```


```julia
julia> plot(zip; components=false)
```

daschw (approved):
Nice, thanks a lot!

src/distributions.jl
```diff
@@ -3,11 +3,11 @@
 function default_range(dist::Distribution, alpha = 0.0001)
     minval = isfinite(minimum(dist)) ? minimum(dist) : quantile(dist, alpha)
     maxval = isfinite(maximum(dist)) ? maximum(dist) : quantile(dist, 1-alpha)
-    minval, maxval
+    promote(minval, maxval)
 end
 
 function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
-    minval = maxval = 0.0
+    minval = maxval = 0
```
devmotion, line 10: I guess this is more likely to be type unstable than the current implementation?

src/distributions.jl
```diff
@@ -18,21 +18,34 @@ function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
...
+    else  # unbounded
+        return (UnitRange(default_range(dist)...),)
+    end
+end
 
 # this "user recipe" adds a default x vector based on the distribution's μ and σ
 @recipe function f(dist::Distribution)
-    if dist isa Distribution{Univariate,Discrete}
-        seriestype --> :scatterpath
+    if dist isa Distribution{Univariate,Discrete} && get(plotattributes, :seriestype, :none) === :none
```
devmotion, line 37: Maybe a bit more standard:
```suggestion
    if dist isa DiscreteUnivariateDistribution && get(plotattributes, :seriestype, :none) === :none
```

src/distributions.jl
```diff
@@ -18,21 +18,34 @@ function default_range(m::Distributions.MixtureModel, alpha = 0.0001)
...
+        seriestype := :sticks
+        markershape --> :circle
     end
     (dist, yz_args(dist)...)
 end
 
 @recipe function f(m::Distributions.MixtureModel; components = true)
+    if m isa Distribution{Univariate,Discrete} && get(plotattributes, :seriestype, :none) === :none
```
devmotion, line 45:
```suggestion
    if m isa DiscreteUnivariateDistribution && get(plotattributes, :seriestype, :none) === :none
```

src/distributions.jl
```diff
@@ -48,8 +61,9 @@ end
     for di in distvec
         @series begin
             seriesargs = isempty(yz) ? yz_args(di) : yz
-            if di isa Distribution{Univariate,Discrete}
-                seriestype --> :scatterpath
+            if di isa Distribution{Univariate,Discrete} && get(plotattributes, :seriestype, :none) === :none
```
devmotion, line 64:
```suggestion
            if di isa DiscreteUnivariateDistribution && get(plotattributes, :seriestype, :none) === :none
```

devmotion (commented):
I wonder if one has to include markers for `:sticks` by default?

src/distributions.jl
sethaxen: Why? It turns out this is necessary for the zero-inflated Poisson to not make everything a `Float64`. Since `logpdf` for discrete `MixtureModel` only accepts points of type `Int`, this turning everything into floats causes failures.

src/distributions.jl
```diff
@@ -3,11 +3,11 @@
 function default_range(dist::Distribution, alpha = 0.0001)
     minval = isfinite(minimum(dist)) ? minimum(dist) : quantile(dist, alpha)
     maxval = isfinite(maximum(dist)) ? maximum(dist) : quantile(dist, 1-alpha)
-    minval, maxval
+    promote(minval, maxval)
```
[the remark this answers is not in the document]
sethaxen: There was, but it was also fixed elsewhere. I've reverted.

src/distributions.jl
devmotion: Because in the loop the type of `minval` and `maxval` will be changed for most distributions: for most distributions `minimum`, `maximum` and `quantile` do not return values of type `Int`. It's also not guaranteed that they are of type `Float64` but this is definitely more likely for continuous distributions (and it would still be type stable for distributions where these values are of type `Int` if one uses `min` and `max` to update `minval` and `maxval` in the loop).

`logpdf` and `pdf` accept `Real` regardless if the distribution is discrete or not. If there's a problem with MixtureModel, it has to be fixed in Distributions.

sethaxen: > Because in the loop the type of `minval` and `maxval` will be changed for most distributions: for most distributions `minimum`, `maximum` and `quantile` do not return values of type `Int`. It's also not guaranteed that they are of type `Float64` but this is definitely more likely for continuous distributions (and it would still be type stable for distributions where these values are of type `Int` if one uses `min` and `max` to update `minval` and `maxval` in the loop).

We should support both though, not just continuous distributions. I pushed an implementation using `mapreduce` that is simpler and does this.

sethaxen:
> I wonder if one has to include markers for `:sticks` by default?

Can you clarify what you mean?

devmotion:
If it is necessary to set `markershape` or if one should just not plot markers with the default settings.

sethaxen:
> If it is necessary to set `markershape` or if one should just not plot markers with the default settings.

Setting `markershape` is necessary to show the markers as suggested in the original issue. Personally, I prefer hair plots without markers and would prefer the simplicity of not explicitly showing them and letting the user do it if they want it. What do you think?

devmotion:
> Personally, I prefer hair plots without markers and would prefer the simplicity of not explicitly showing them and letting the user do it if they want it.

Yes, my question was if you think it would be preferrable to not show them explicitly. I prefer the version without markers (IMO it is a bit cleaner, in particular if there are many values to plot) and it seems simple enough to specify `markershape` if one wants to plot points as well.

sethaxen:
Okay, I simplified the implementation to not show markers by default.

```julia
julia> plot(Binomial(50, 0.5))
```


For plots with few points, in my view this looks worse:

```julia
julia> 
plot(plot(Binomial(5, 0.1)), plot(Binomial(5, 0.1); markershape=:circle))
```


But it looks much better for plots with many points:

```julia
julia> plot(plot(Poisson(500)), plot(Poisson(500); markershape=:circle); legend=:left)
```

BeastyBlacksmith (approved):

PR #453, opened 2021-06-28
Improvements to ecdfplot

sethaxen:
This PR makes several improvements to `ecdfplot`:
- allow setting `seriestype` to something besides `steppost`.
- if `seriestype` is not `steppost`, don't plot the extra point on the x-axis.
- don't plot non-unique points. e.g., when plotting the ECDF of a `Binomial(50, 0.1)` sample, at most 51 points should be plotted, but previously it plotted `length(x)` points.
- allow specifying maximum number of points to plot, which is important when the sample is very large.
- add a docstring

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StatsPlots"
 uuid = "f3b207a7-027a-5e70-b257-86293d7955fd"
-version = "0.14.23"
+version = "0.14.24"
 
 [deps]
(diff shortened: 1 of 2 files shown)


sethaxen:
I'd also like to add the ability to provide the evaluation points as `x`, so one could do `ecdfplot(0:0.1:1, rand(100))`.

sethaxen:
Before this is merged, I'd like to add the ability to specify the evaluation points.

PR #465, opened 2021-09-08
Fixes normalization for groupedhist

### LINKED ISSUE #457: [BUG] Normalization not working for `groupedhist`
If `group`s are used the bins are no longer normalized:
```julia
StatsPlots.groupedhist(
    [randn(1000) randn(1000) .+ 2.0], 
    group=repeat(["1", "2"], inner=1000), 
    bar_position=:doge, 
    normalize=true,
)
```
bug

ValentinKaisermayer:
fixes #457

### DIFF
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -98,7 +98,11 @@ Plots.group_as_matrix(g::GroupedHist) = true
             v_i = filter(x->!isnan(x), v[:,i])
             w_i = weights == nothing ? nothing : weights[groupinds]
             h_i = Plots._make_hist((v_i,), h.edges; normed = false, weights = w_i)
-            y[:,i] = normed ? (h_i.weights / sum(h_i.weights)) .* (length(v_i) / ntot) : h_i.weights
+            if normed
+                y[:,i] .= h_i.weights .* (length(v_i) / ntot / sum(h_i.weights))
+            else
+                y[:,i] .= h_i.weights
+            end
         end
     end
 
--- a/src/hist.jl
+++ b/src/hist.jl
@@ -97,7 +97,7 @@ Plots.group_as_matrix(g::GroupedHist) = true
             v_i = filter(x->!isnan(x), v[:,i])
             w_i = weights == nothing ? nothing : weights[groupinds]
             h_i = Plots._make_hist((v_i,), h.edges; normed = false, weights = w_i)
-            y[:,i] = normed ? (h_i.weights / sum(h_i.weights)) .* (length(groupinds) ./ length(group)) : h_i.weights
+            y[:,i] = normed ? (h_i.weights / sum(h_i.weights)) .* (length(v_i) / ntot) : h_i.weights
         end
     end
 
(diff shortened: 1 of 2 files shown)


src/hist.jl
```diff
@@ -97,7 +97,7 @@ Plots.group_as_matrix(g::GroupedHist) = true
             v_i = filter(x->!isnan(x), v[:,i])
             w_i = weights == nothing ? nothing : weights[groupinds]
             h_i = Plots._make_hist((v_i,), h.edges; normed = false, weights = w_i)
-            y[:,i] = normed ? h_i.weights / sum(h.weights) : h_i.weights
+            y[:,i] = normed ? (h_i.weights / sum(h_i.weights)) .* (length(groupinds) ./ length(group)) : h_i.weights
```
sethaxen, line 100: The only thing this is missing is handling `NaN`s correctly. `Plots._make_hist` removes `NaN`'s (see https://github.com/JuliaPlots/Plots.jl/blob/v1.22.3/src/recipes.jl#L823), hence why `v_i` filters `NaN`s. So we instead need the numerator to be the length of `v_i`, and the denominator to be the number of non-`NaN`s in `v`, i.e. `ntot = count(x->!isnan(x), v)`.

Then,

```suggestion
            y[:,i] = normed ? (h_i.weights / sum(h_i.weights)) .* (length(v_i) / ntot) : h_i.weights
```

src/hist.jl
```diff
@@ -82,6 +82,7 @@ Plots.group_as_matrix(g::GroupedHist) = true
     edges = h.edges[1]
     bar_width --> mean(map(i -> edges[i+1] - edges[i], 1:nbins))
     x = map(i -> (edges[i] + edges[i+1])/2, 1:nbins)
+    ntot = count(x->!isnan(x), v)
```
sethaxen, line 85: Since this is only needed if `group !== nothing`, can you moved it to below the `else`?

src/hist.jl
```diff
@@ -97,7 +98,7 @@ Plots.group_as_matrix(g::GroupedHist) = true
             v_i = filter(x->!isnan(x), v[:,i])
             w_i = weights == nothing ? nothing : weights[groupinds]
             h_i = Plots._make_hist((v_i,), h.edges; normed = false, weights = w_i)
-            y[:,i] = normed ? h_i.weights / sum(h.weights) : h_i.weights
+            y[:,i] = normed ? (h_i.weights / sum(h_i.weights)) .* (length(v_i) / ntot) : h_i.weights
```
sethaxen, line 101: This eliminates two allocations:
```suggestion
            if normed
                y[:,i] .= h_i.weights .* (length(v_i) / ntot / sum(h_i.weights))
            else
                y[:,i] .= h_i.weights
            end
```

sethaxen (changes requested):
Just a couple of performance tweaks, then I think this is ready to merge.

sethaxen:
Can you also increment the patch version number?

sethaxen:
Thanks @ValentinKaisermayer for the PR!

PR #468, opened 2021-10-22
Support drawing boxplots horizontally

tpgillam:
Addresses https://github.com/JuliaPlots/StatsPlots.jl/issues/251

Inspired by the way that the `orientation` attribute is handled in the `bar` recipe inside `Plots.jl`


tpgillam:
Just wondering if there's anything else I can do on this? Thanks!

mkborregaard:
Thanks!

PR #473, opened 2021-12-07
Fix boxplot `range` attribute by renaming it `whisker_range`

yha:
`range` is now an alias to `lims` (JuliaPlots/Plots.jl#3803), so it can't be used in the recipe.
Docs update in a separate PR to `Plots`.


diegozea:
Thanks, this should fix the issue that I have just noticed of the non-working `range` attribute:

image

diegozea:
Thanks! Is it possible to release a new version with this fix?

yha:
https://github.com/JuliaRegistries/General/pull/51008

PR #480, opened 2022-01-26
wand bins patch: fix #479

### LINKED ISSUE #479: MethodError (pdf is ambiguous) when using wand bins (histogram)
The following is an example of the error:

Screenshot from 2021-12-10 13-42-20

You can reproduce it using:

```julia
import Pkg; Pkg.activate(temp=true); Pkg.add(["StatsPlots", "RDatasets"])
using StatsPlots, RDatasets
@df iris histogram(:PetalLength, bins=:wand)
```

The error is:

```julia
julia> @df iris histogram(:PetalLength, bins=:wand)
ERROR: MethodError: pdf(::Distributions.Normal{Float64}, ::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}}) is ambiguous. Candidates:
  pdf(d::Distributions.Distribution{Distributions.ArrayLikeVariate{N}, S} where S<:Distributions.ValueSupport, x::AbstractArray{var"#s11", M} where var"#s11"<:Real) where {N, M} in Distributions at /home/diego/.julia/packages/Distributions/hnQlR/src/common.jl:249
  pdf(d::Distributions.UnivariateDistribution{S} where S<:Distributions.ValueSupport, X::AbstractArray) in Distributions at deprecated.jl:70
Possible fix, define
  pdf(::Distributions.UnivariateDistribution{S} where S<:Distributions.ValueSupport, ::AbstractArray{T, M} where T<:Real) where M
Stacktrace:
  [1] bkfe(gcounts::Vector{Float64}, drv::Int64, bandwidth::Float64, range_x::Vector{Float64})
    @ StatsPlots ~/.julia/packages/StatsPlots/dFNaF/src/hist.jl:215
  [2] wand_bins(x::Vector{Float64}, scalest::Symbol, gridsize::Int64, range_x::Tuple{Float64, Float64}, trun::Bool)
    @ StatsPlots ~/.julia/packages/StatsPlots/dFNaF/src/hist.jl:147
  [3] wand_bins(x::Vector{Float64})
    @ StatsPlots ~/.julia/packages/StatsPlots/dFNaF/src/hist.jl:122
  [4] wand_edges(::Vector{Float64})
    @ StatsPlots ~/.julia/packages/StatsPlots/dFNaF/src/hist.jl:117
  [5] _auto_binning_nbins(vs::Tuple{Vector{Float64}}, dim::Int64; mode::Symbol)
    @ Plots ~/.julia/packages/Plots/FCM0H/src/recipes.jl:798
  [6] _hist_edge
    @ ~/.julia/packages/Plots/FCM0H/src/recipes.jl:806 [inlined]
  [7] JuliaPlots/StatsPlots.jl#222
    @ ~/.julia/packages/Plots/FCM0H/src/recipes.jl:817 [inlined]
  [8] map
    @ ./tuple.jl:213 [inlined]
  [9] _hist_edges
    @ ~/.julia/packages/Plots/FCM0H/src/recipes.jl:817 [inlined]
 [10] _make_hist(vs::Tuple{Vector{Float64}}, binning::Symbol; normed::Bool, weights::Nothing)
    @ Plots ~/.julia/packages/Plots/FCM0H/src/recipes.jl:836
 [11] macro expansion
    @ ~/.julia/packages/Plots/FCM0H/src/recipes.jl:860 [inlined]
 [12] apply_recipe(plotattributes::AbstractDict{Symbol, Any}, #unused#::Type{Val{:barhist}}, x::Any, y::Any, z::Any)
    @ Plots ~/.julia/packages/RecipesBase/qpxEX/src/RecipesBase.jl:289
 [13] _process_seriesrecipe(plt::Any, plotattributes::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/Bxu2O/src/series_recipe.jl:50
 [14] _process_seriesrecipe(plt::Any, plotattributes::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/Bxu2O/src/series_recipe.jl:60
 [15] _process_seriesrecipes!(plt::Any, kw_list::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/Bxu2O/src/series_recipe.jl:27
 [16] recipe_pipeline!(plt::Any, plotattributes::Any, args::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/Bxu2O/src/RecipesPipeline.jl:97
 [17] _plot!(plt::Plots.Plot, plotattributes::Any, args::Any)
    @ Plots ~/.julia/packages/Plots/FCM0H/src/plot.jl:208
 [18] #plot#165
    @ ~/.julia/packages/Plots/FCM0H/src/plot.jl:91 [inlined]
 [19] histogram(args::Any; kw::Any)
    @ Plots ~/.julia/packages/RecipesBase/qpxEX/src/RecipesBase.jl:410
 [20] add_label(argnames::Vector{String}, f::Function, args::Vector{Float64}; kwargs::Base.Iterators.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:bins,), Tuple{Symbol}}})
    @ StatsPlots ~/.julia/packages/StatsPlots/dFNaF/src/df.jl:153
 [21] (::var"#1#2")(257::DataFrame)
    @ Main ./none:0
 [22] top-level scope
    @ REPL[5]:1
```

The version info is:

```julia
(jl_Ow3p6i) pkg> st
      Status `/tmp/jl_Ow3p6i/Project.toml`
  [ce6b1742] RDatasets v0.7.6
  [f3b207a7] StatsPlots v0.14.29

julia> versioninfo()
Julia Version 1.6.2
Commit 1b93d53fc4 (2021-07-14 15:36 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)

```

mkborregaard:
Seemed generally bitrotted. Missing a dep for the FFT, the pdf function was changed in Distributions long ago. For wand bins to actually work also requires https://github.com/JuliaPlots/Plots.jl/pull/4076 in Plots (version XX)


sethaxen:
I'd prefer we just used the [AbstractFFTs](https://github.com/JuliaMath/AbstractFFTs.jl) interface. Once one of the packages that implements the interface is loaded, there's no way for the user to switch between them, and the whole point of the interface package is to not coerce the user to a choice. So we can document that to use wand bins, the user should load an FFT package  (e.g. FFTW or FastTransforms). And if one of the other packages in their environment already did, then great, it already works.

mkborregaard:
Are you sure? That sounds strange, why would the user not be able to choose a different FFT than the one we use internally here? Not really keen on hoisting the dep management manually onto the user.

sethaxen:
> Are you sure? That sounds strange, why would the user not be able to choose a different FFT than the one we use internally here?

Currently only if the user constructs their own FFT plan can they coerce the backend. The `fft` function takes no backend argument, so it ends up using one of the packages that has been loaded but the user can't control which. See https://github.com/JuliaMath/AbstractFFTs.jl/issues/32

> Not really keen on hoisting the dep management manually onto the user.

I think this is standard for FFT though. I'll look into it some more.

mkborregaard:
So the FFTW dep is a bit strange to me - this functionality obviously worked when it was added, so `fft` must have been in the namespace without FFTW at that time. What package defined that?

piever:
We depend on KernelDensity for the smooth density plots, and they load FFTW ([see here](https://github.com/JuliaStats/KernelDensity.jl/blob/922dfffb99d54eb1f9f88f778d1ac44eb0dcc738/src/KernelDensity.jl#L10)), so even if we get the function from AbstractFFT (which I guess we might), there will already be the FFTW backend loaded due to KernelDensity.

I personally would also prefer it if KernelDensity was FFTW backend agnostic (due to the MKL dependency of FFTW.jl), but 
1. that should be probably discussed over at KernelDensity.jl, and
2. it would IMO only really make sense once there is more than one implementation of FFT for cpu arrays.

sethaxen:
> So the FFTW dep is a bit strange to me - this functionality obviously worked when it was added, so `fft` must have been in the namespace without FFTW at that time. What package defined that?

According to the [AbstractFFTs docs](https://juliamath.github.io/AbstractFFTs.jl/stable/), it used to be in the stdlib for Julia 0.6 and earlier. StatsPlots dropped support for v0.6 in [v0.8.0](https://github.com/JuliaPlots/StatsPlots.jl/releases/tag/v0.8.0). Wand bins were added in v0.5.0.

sethaxen:
> We depend on KernelDensity for the smooth density plots, and they load FFTW ([see here](https://github.com/JuliaStats/KernelDensity.jl/blob/922dfffb99d54eb1f9f88f778d1ac44eb0dcc738/src/KernelDensity.jl#L10)), so even if we get the function from AbstractFFT (which I guess we might), there will already be the FFTW backend loaded due to KernelDensity.

Ah okay, well that then makes this discussion moot. But I would prefer to just depend on AbstractFFTs directly then.

mkborregaard:
@sethaxen I understand your concern as it seems we are adding a new binary dep, but actually KernelDensity here already depends directly on FFTW, so there is no difference in terms of footprint. If you prefer, I can do `import KernelDensity.FFTW: fft`? Though I think it is better practice to make the dep explicit

mkborregaard:
@sethaxen alright, I've updated the dep to AbstractFFTs. Seems to work perfectly. I also added a test that isn't numerically precise (because one shouldn't assume random numbers in tests) but just sees if the function runs essentially.
For some reason the Distribution tests fail on my machine when I run the StatsPlots tests locally, but that is orthogonal.

src/hist.jl
```diff
@@ -207,12 +207,12 @@ function bkfe(gcounts, drv, bandwidth, range_x)
     ## Obtain kernel weights
 
     tau = 4 + drv
-    L = min(floor(tau*h/delta), M)
+    L = min(floor(Int, tau*h/delta), M)
```
sethaxen, line 210:
```suggestion
    L = min(Int(fld(tau*h, delta)), M)
```

src/hist.jl
```diff
@@ -207,12 +207,12 @@ function bkfe(gcounts, drv, bandwidth, range_x)
...
 
     lvec = 0:L
     arg = lvec .* delta/h
 
-    kappam = pdf(Normal(),arg)/(h^(drv+1))
+    kappam = pdf.(Normal(),arg)/(h^(drv+1))
```
sethaxen, line 215:
```suggestion
    kappam = pdf.(Normal(),arg) ./ h^(drv+1)
```

src/hist.jl
```diff
@@ -227,7 +227,7 @@ function bkfe(gcounts, drv, bandwidth, range_x)
 
     ## Now combine weights and counts to obtain estimate
     ## we need P >= 2L+1L, M: L <= M.
-    P = 2^(ceil(log(M+L+1)/log(2)))
+    P = 2^(ceil(Int, log(M+L+1)/log(2)))
```
sethaxen, line 230: Looks like this is just change of base to 2:
```suggestion
    P = 2^ceil(Int, log2(M+L+1))
```

src/StatsPlots.jl
```diff
@@ -20,7 +20,7 @@ import Clustering: Hclust, nnodes
 using Interpolations
 import MultivariateStats: MDS, eigvals, projection, principalvars,
                             principalratio, transform
-
+import AbstractFFTs: fft, ifft
```
sethaxen, line 23: Importing implies we overload them, which we don't.
```suggestion
using AbstractFFTs: fft, ifft
```

sethaxen (changes requested):
A few minor changes

src/hist.jl
devmotion: It seems this is just
```suggestion
    P = nextpow(2, M+L+1)
```

src/StatsPlots.jl
mkborregaard: I agree, though importing anyway is the convention here.

src/hist.jl
mkborregaard: Why is that better? I've commited your suggestion but don't really understand it.

mkborregaard:
Changed as requested.

sethaxen (approved):
LGTM!

PR #483, opened 2022-02-02
add feature to change x-axis order for boxplots

### LINKED ISSUE #409: Expose sorting options for boxplots (and other plots?)
By default the boxplot orders a categorical axis alphabetically.  Often times, there is a semantic ordering that makes more sense.  Allowing the user to specify the order for the x-axis would be a meaningful enhancement.

Tobias-Thomas:
Fixes #409, by adding `by` as a new optional parameter to boxplots. This `by` is the same, one would use for `sort`.  It's default parameter is the `identity`.
Before merge, can someone please tell me, where I can add the documentation for this new parameter?
Thanks in advance.


sethaxen:
@Tobias-Thomas thank you for the contribution, and I'm very sorry for the late review! I know I recommended the keyword `by`, but in retrospect, I think that's too generic of a name. What about something like `sort_labels_by`, or perhaps you have a better suggestion?

Tobias-Thomas:
Oh, no problem. The only other idea, I would have is `sort_x_by`, since the parameter it sorts is called `x`.
What do you think?

indymnv:
I was searching for a way to sort boxplots, but still is not solved or I am wrong?, is there a way I can help to fix this?

Tobias-Thomas:
The fix I implemented was essentially quite easy, but as you can see above, we were still unsure about the name of this parameter, and I think we need a JuliaPlots member to decide on that.

indymnv:
What about order to keep a similar idea like [seaborn](https://stackoverflow.com/questions/59245378/python-seaborn-sorting-based-on-y-axis-instead-of-x-axis), sort_labels_by also can be OK, sort_x_by can be a little bit confused if later there is an implementation using horizontal boxplots.

Tobias-Thomas:
Yea, I understand that point. Out of those options, I would `sort_labels_by` prefer.
I will do a commit in my branch (switching to this name) later this day and maybe we could ping Seth again, to see if he approves.

Tobias-Thomas:
@sethaxen, I updated the name and merged it into the current master. Is there anything left to do for me in this PR?

Tobias-Thomas:
I think the fix would be ready for merge, but I do not know who to ping, to review the PR..

indymnv:
Maybe @sethaxen , can help us with this PR

src/boxplot.jl
```diff
@@ -31,7 +32,7 @@ notch_width(q2, q4, N) = 1.58 * (q4 - q2) / sqrt(N)
     isnothing(bw) && (bw = 0.8)
     @assert whisker_width === :match || whisker_width == :half || whisker_width >= 0 "whisker_width must be :match, :half, or a positive number"
     ww = whisker_width === :match ? bw : whisker_width == :half ? bw / 2 : whisker_width
-    for (i, glabel) in enumerate(glabels)
+    for (i, glabel) in enumerate(sort(glabels, by = sort_labels_by))
```
sethaxen, line 35:
```suggestion
    for (i, glabel) in enumerate(sort(glabels; by = sort_labels_by))
```

sethaxen (approved):
Sorry @Tobias-Thomas for the delay! Everything looks good to me! Thanks!

Vaeryus:
So how do I use sort_labels_by now. Can I just put in an order for the x axis like:
 sort_labels_by= [ "B" "C" "A"] or how would I use it?

Tobias-Thomas:
> So how do I use sort_labels_by now? Can I just put in an order for the x axis like: sort_labels_by= [ "B" "C" "A"] or how would I use it?

No, the function uses the same interface as `Base.sort`. That means that values can be sorted by a transformation of their values. So you need to write a function which, transforms the values into a space where they are sorted the way you want it.

artkel:
Can someone please explain how the category sorting for boxplots now works? It is not quite clear to me from the previous comment. Would be nice to see an example. Thanks!

mkborregaard:
Before this PR the way to do that was to use `levels!` on the CategoricalArray x vector to define the sorting order. Is this really an improvemet on that?

PR #491, opened 2022-02-21
Add coefplot and groupedcoefplot for StatisticalModel

getzze:
The new dependency is minimal (StatsModels), test dependencies are a bit bigger (CategoricalArrays, DataFrames, GLM).
Tested with the gr, pyplot and plotly backends.
It was tested and working with GLM.jl, MixedModels.jl and RobustModels.jl models. I don't know other packages that use the StatsModels API.


README.md
```diff
@@ -494,3 +494,23 @@ covellipse([0,2], [2 1; 1 4], n_std=2, aspect_ratio=1, label="cov1")
 covellipse!([1,0], [1 -0.5; -0.5 3], showaxes=true, label="cov2")
 ```
 ![covariance ellipses](https://user-images.githubusercontent.com/4170948/84170978-f0c2f380-aa82-11ea-95de-ce2fe14e16ec.png)
+
+## Coefficient plot
+
+Coefficients of a RegressionModel can be plotted with their standard deviation to easily visualize effect size.
```
sethaxen, line 500:
```suggestion
Coefficients of a `RegressionModel` can be plotted with their standard deviation to easily visualize effect size.
```

sethaxen: Can you also prefix `RegressionModel` with either the package that defines the type (`StatsAPI`) or the package that implements the code that returns the type?

src/statsmodels.jl
```diff
@@ -0,0 +1,549 @@
+
+
+### Util functions
+
+StatsBase.responsename(m) = coefnames(formula(m).lhs)
```
sethaxen, line 5: I'd prefer we work with StatsAPI directly:
```suggestion
StatsAPI.responsename(m) = StatsAPI.coefnames(StatsAPI.formula(m).lhs)
```

src/statsmodels.jl
```diff
@@ -0,0 +1,549 @@
...
+
+function baselevel(ct::CategoricalTerm)
```
sethaxen, line 7: Can you prefix `CategoricalTerm` with the package that defines the type for clarity? Also for the below method definitions.

sethaxen (changes requested):
Thanks for the PR! I'm happy to have this functionality in this package. It's a lot of code though, so we'll probably need a few rounds of review.

A few general comments first:
- I know StatsPlots doesn't consistently do this (yet), but it would make the code more readable if the function calls/types were prefixed with the package name where they're defined (or implemented, if we rely on the implementation)
- Can you make your utility function names more descriptive? This will make it easier for me to follow the code.
- You've essentially implemented a forest plot. This is generally useful apart from plotting coefficients of statistical models. Would it make sense to refactor to a `forestplot` primitive and a `coefplot` that extracts the coefficients and passes them to `forestplot`?

getzze:
Hi, sorry for the long delay. I had some time so I refactored the code, let me know if it is clearer.
The util functions to parse the formula terms are very specific so I didn't comment them fully.

PR #492, opened 2022-03-25
Add support for horizontal violin and dot plots

churli:
The horizontal::Bool kwarg has been added to the violin and dotplot
recipes and this allows generating horizontal versions of these plots.

### DIFF
--- a/src/dotplot.jl
+++ b/src/dotplot.jl
@@ -2,7 +2,11 @@
 # ---------------------------------------------------------------------------
 # Dot Plot (strip plot, beeswarm)
 
-@recipe function f(::Type{Val{:dotplot}}, x, y, z; mode = :density, side=:both)
+@recipe function f(::Type{Val{:dotplot}}, x, y, z; 
+                        mode = :density, 
+                        side=:both,
+                        horizontal=false,
+                        )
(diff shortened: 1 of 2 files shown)


churli:
For context: I needed to have horizontal violin e boxplots and found it easy to tweak the StatsPlots library to get them, now I'm sharing this back with you.
Cheers!

BeastyBlacksmith:
That looks good already, it would be good to use the existing attribute `orientation` though. Maybe you can compare to the recipe for boxplots ( that got this in https://github.com/JuliaPlots/StatsPlots.jl/commit/be7f9c03110242707915aaaae8e1e316b179a3d5).

churli:
Thanks a lot for the suggestion, my bad that I didn't check how it was implemented for boxplots.
Now it works through the orientation attribute for violins and dotplots as well.

BeastyBlacksmith:
It doesn't account for the tick labels though.
Here is a MWE
```julia
using StatsPlots
import RDatasets
singers = RDatasets.dataset("lattice", "singer")
@df singers violin(:VoicePart, :Height, line = 0, fill = (0.2, :blue), orientation = :horizontal)
```
violins

PR #513, opened 2022-08-22
add errorline.jl

CMGreenspon:
Function for parsing inputs to easily make a ribbons (https://ggplot2.tidyverse.org/reference/geom_ribbon.html)  or errorbar https://www.mathworks.com/help/matlab/ref/errorbar.html) plot while allowing for easily controlling error type and also handles NaNs.


.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
BeastyBlacksmith, line 1: This shouldn't be here

src/errorline.jl
```diff
@@ -0,0 +1,179 @@
...
+
+# Inputs: default values are indicated with *s
+
+    x (vector, unit range) - the values along the x-axis for each y-point
+
+    y (matrix [x, repeat, group]) - values along y-axis wrt x. The first dimension must be of equal length to that of x.
+        The second dimension is treated as the repeated observations and error is computed along this dimension. If the 
+        matrix has a 3rd dimension this is treated as a new group.
+
+    ErrorStyle (symbol - *:Ribbon* or :Classic) - determines whether to use a ribbon style or stick style error representation.
```
BeastyBlacksmith, line 16:
```suggestion
    error_style (symbol - *:ribbon* or :classic) - determines whether to use a ribbon style or stick style error representation.
```
In julia usually only types and modules have upper CamelCase

src/errorline.jl
```diff
@@ -0,0 +1,179 @@
...
+    y = fill(NaN, 10, 100, 3)
+    for i = axes(y,3)
+        y[:,:,i] = collect(1:2:20) .+ rand(10,100).*5 .* collect(1:2:20) .+ rand()*100
+    end
+    errorline(1:10, y)
+    ```
+"""
+errorline
+
+function compute_error(y::AbstractMatrix, CenterType::Symbol, ErrorType::Symbol, Percentiles::Vector{Int64})
```
BeastyBlacksmith, line 45:
```suggestion
function compute_error(y::AbstractMatrix, CenterType::Symbol, ErrorType::Symbol, Percentiles::Vector{Int})
```
This currently fails the tests on 32bit machines. Any reason this couldn't be `percentiles::AbstractVector`?

src/errorline.jl
```diff
@@ -0,0 +1,179 @@
...
+
+        # Check y orientation
+        if ndims(y) > 3
+            error("ndims(y) > 3")
+        end
+
+        if !any(size(y) .== length(x))
+            error("Size of x and y do not match")
+        elseif ndims(y) == 2 && size(y,1) != length(x) && size(y,2) == length(x) # Check if y needs to be transposed or transmuted
+            y = y'
```
BeastyBlacksmith, line 100: `y'` is the adjoint you might want to use `transpose` or `permutedims` here

src/errorline.jl
```diff
@@ -0,0 +1,179 @@
...
+        elseif ndims(y) == 3 && size(y,1) != length(x) 
+            error("When passing a 3 dimensional matrix as y, the axes must be [x, repeat, group]")
+        end
+    end
+
+    # Parse different color type
+    if typeof(GroupColor) == Symbol || typeof(GroupColor) == RGB{Float64}
```
BeastyBlacksmith, line 107:
```suggestion
    if GroupColor isa Symbol || GroupColor isa RGB{Float64}
```
What about `RGBA`?

BeastyBlacksmith (changes requested):
Thank you for your contribution. I got some stylistic and functional remarks.
Additionally we should think about whether this recipe needs to live here or could also be added to `Plots.jl` itself.
Currently, I don't see a reason why not.

src/errorline.jl
BeastyBlacksmith: Same applies for the other keyword arguments

CMGreenspon:
I believe I've addressed all the changes. The error you got was because NaNMath only accepts Floats so I've added a conversion check. I also added the final distribution mode.

BeastyBlacksmith (approved):
So far this looks fine to me, there should be at least one example in the README as Documentation. Would be good to have another pair of eyes here.

CMGreenspon:
I'll add an example or two to the readme in an hour or so

src/errorline.jl
```diff
@@ -0,0 +1,236 @@
...
+    errortype (symbol - *:std*, :sem, :percentile) - which error metric to use to show the distribution of y at each x-value.
+
+    percentiles (Vector{Int64} *[25, 75]*) - if using errortype == :percentile then which percentiles to use as bounds.
+
+    groupcolor (Symbol, RGB, Vector of Symbol or RGB) - Declares the color for each group. If no value is passed then will use
+        the default colorscheme. If one value is given then it will use that color for all groups. If multiple colors are
+        given then it will use a different color for each group.
+
+    secondarycolor (Symbol, RGB, :matched - *:Gray60*) - When using stick mode this will allow for the setting of the stick color.
+        If ":matched" is given then the color of the sticks with match that of the main line.
```
t-bltg, line 31:
```suggestion
        If `:matched` is given then the color of the sticks with match that of the main line.
```

Use backquotes for references to keywords.

src/errorline.jl
```diff
@@ -0,0 +1,236 @@
...
+
+    errortype (symbol - *:std*, :sem, :percentile) - which error metric to use to show the distribution of y at each x-value.
+
+    percentiles (Vector{Int64} *[25, 75]*) - if using errortype == :percentile then which percentiles to use as bounds.
+
+    groupcolor (Symbol, RGB, Vector of Symbol or RGB) - Declares the color for each group. If no value is passed then will use
+        the default colorscheme. If one value is given then it will use that color for all groups. If multiple colors are
+        given then it will use a different color for each group.
+
+    secondarycolor (Symbol, RGB, :matched - *:Gray60*) - When using stick mode this will allow for the setting of the stick color.
```
t-bltg, line 30:
```suggestion
    secondarycolor (`Symbol`, `RGB`, `:matched` - *:Gray60*) - When using stick mode this will allow for the setting of the stick color.
```

src/errorline.jl
```diff
@@ -0,0 +1,236 @@
...
+
+# Inputs: default values are indicated with *s
+
+    x (vector, unit range) - the values along the x-axis for each y-point
+
+    y (matrix [x, repeat, group]) - values along y-axis wrt x. The first dimension must be of equal length to that of x.
+        The second dimension is treated as the repeated observations and error is computed along this dimension. If the
+        matrix has a 3rd dimension this is treated as a new group.
+
+    error_style (symbol - *:ribbon*, :stick, :spaghetti) - determines whether to use a ribbon style or stick style error
```
t-bltg, line 17:
```suggestion
    error_style (`Symbol` - *:ribbon*, :stick, :spaghetti) - determines whether to use a ribbon style or stick style error
```

src/errorline.jl
```diff
@@ -0,0 +1,236 @@
...
+                    linecolor := groupcolor[g]
+                end
+                ()
+            end
+
+        elseif errorstyle == :spaghetti
+            num_obs = size(y,2)
+            if num_obs > numsecondarylines
+                sub_sample_idx = sample(1:num_obs, numsecondarylines, replace=false)
+                y_sub_sample = y[:,sub_sample_idx,g]
```
t-bltg, line 196: I don't know what convention is used in `StatsPlots` (maybe we need an autoformatter here as well), but I find `y[:, sub_sample_idx, g]` more readable.

t-bltg (commented):
I'm not a `StatsPlots` user so I'm only going to add stylistic comments here. Otherwise, I don't see why it cannot go in, since this PR is tested and documented.

src/errorline.jl
```diff
@@ -0,0 +1,236 @@
...
+                fillalpha --> .1
+                if groupcolor !== nothing
+                    linecolor := groupcolor[g]
+                    fillcolor := groupcolor[g]
+                end
+                () # Supress implicit return
+            end
+
+        elseif errorstyle == :stick
+            x_offset = (extrema(x)[2] - extrema(x)[1]) * stickwidth
```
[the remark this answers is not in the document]
CMGreenspon: Thank you for this! I come from matlab and googling range in julia didn't give me what I wanted to this is perfect!

src/errorline.jl
```diff
@@ -0,0 +1,236 @@
...
+    # NaNMath doesn't accept Ints so convert to AbstractFloat if necessary
+    if eltype(y) <: Integer
+        y = float(y)
+    end
+    # First compute the center
+    if centertype == :mean
+        y_central =  mapslices(NaNMath.mean, y, dims=2)
+    elseif centertype == :median
+        y_central =  mapslices(NaNMath.median, y, dims=2)
+    end
```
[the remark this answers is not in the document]
CMGreenspon: I don't love that line 60 is a single line, I think it makes it awkward to read, but if broken into two lines I think that's pretty good so I'll do that unless that's against convention?

CMGreenspon:
plot_36
Example plot

CMGreenspon:
ExamplePlot
Rename spaghetti to plume

CMGreenspon:
ExamplePlot

PR #518, opened 2022-09-06
Fix secondary color

CMGreenspon:
Reworked how the secondary color is computed and fixed typo in readme

Fix: #517

### DIFF
--- a/src/errorline.jl
+++ b/src/errorline.jl
@@ -117,21 +117,26 @@ end
         end
     end
 
-    # Parse different color type
-    if groupcolor isa Symbol || groupcolor isa RGB{Float64} || groupcolor isa RGBA{Float64}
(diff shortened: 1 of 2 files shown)


src/errorline.jl
```diff
@@ -141,17 +146,19 @@ end
     for g = axes(y,3) # Iterate through 3rd dimension
         # Compute center and distribution for each value of x
         y_central, y_error = compute_error(y[:,:,g], centertype, errortype, percentiles)
+
+        # Check secondarycolor validity
+        temp = 0;
```
BeastyBlacksmith, line 151: Where is that varaible been used?

CMGreenspon: oh whoops, left that in during debugging, 2 seconds

PR #519, opened 2022-09-08
add a formatter check - initial formatting (nfc)

t-bltg:
1. non functional change: indentation is inconsistent as seen in https://github.com/JuliaPlots/Plots.jl/pull/4339;
2. update workflows.

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,13 @@
+annotate_untyped_fields_with_any = false
+short_to_long_function_def = false
+long_to_short_function_def = true
+whitespace_ops_in_indices = true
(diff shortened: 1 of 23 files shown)


.JuliaFormatter.toml
```diff
@@ -0,0 +1,13 @@
+annotate_untyped_fields_with_any = false
+short_to_long_function_def = false
+long_to_short_function_def = true
```
BeastyBlacksmith, line 3: I object here. If I write a long form function, I usually have a reason for that and don't want the formatter to override that

t-bltg: It is conditional: it respect the margins, and converts only if there a single statement.

Issue #523, opened 2023-02-02
errorline! not cycling palette


alancummings:
If I run the sample code at https://docs.juliaplots.org/latest/generated/statsplots/#ErrorLine the colors do not cycle correctly.

- The first plot is blue (appropriate)
- The second plot is orange (appropriate)
- The 3rd plot is orange again (**NOT** appropriate)

**To reproduce**: I start with a fresh julia REPL  `using StatsPlots` and copy paste the example from the link above

**Additional info**: I modified the example to create a plot with `p = plot()`, then use `errorline!` for all 3 series. In this case **all 3 plots are _orange_**

I am running julia version 1.8.5 in a VSCode type environment (under code-server: https://github.com/coder/code-server) on an AWS amazon linux 2 instance

BeastyBlacksmith:
CC: @CMGreenspon

CMGreenspon:
Okay, it seems like the GR backend changed how it counts the number of primary series. I've implemented a fix to make it explicitly count the number of primary series instead of relying on the GR counter.

BeastyBlacksmith:
Would be good to know on which version of Plots this behaved differently. I don't think it is related to GR

alancummings:
> 

I have Plots v1.38.3 and StatsPlots v0.15.4

PR #524, opened 2023-02-03
Error line pallete bug fix

CMGreenspon:
Explicitly counts number of primary series as the length of the series object no longer seems to track that

### DIFF
--- a/src/errorline.jl
+++ b/src/errorline.jl
@@ -152,9 +152,9 @@ end
                 gsi_counter += 1
             end
         end
-        # Allow wrapping of indices
-        group_series_index = gsi_counter + 1
-        idx = (group_series_index:group_series_index+size(y,3)) .% length(palette(color_palette))
+        # Start at next index and allow wrapping of indices
+        gsi_counter += 1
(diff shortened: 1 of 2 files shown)


CMGreenspon:
Closes #523

src/errorline.jl
```diff
@@ -50,45 +50,45 @@ errorline(1:10, y)
...
+    if errortype === :std || errortype === :sem
+        y_error = mapslices(NaNMath.std, y, dims=2)
+        if errortype == :sem
+            y_error = y_error ./ sqrt(size(y,2))
         end
 
-    elseif errortype === :percentile
-        y_lower = fill(NaN, size(y, 1))
-        y_upper = fill(NaN, size(y, 1))
+    elseif errortype == :percentile
```
BeastyBlacksmith, line 80: would be good to consistently use `===` for comparisons with symbols

BeastyBlacksmith:
Can you add a test that checks that you get the correct color for the example in #523 ?

CMGreenspon:
I can add the check but I have a bunch of deadlines before I go on holiday for a week so it will have to wait until the week of the 20th.

PR #535, opened 2023-04-26
Remove dependency on DataValues.jl

svilupp:
Closes https://github.com/JuliaPlots/StatsPlots.jl/issues/534

All tests pass locally. I have not found any impacted functionality.


sethaxen:
Makes sense to me. #87 introduced this dependency because it was already a dependency of IterableTables, which this package no longer depends on, so I suspect this can be removed. @mkborregaard can you think of a reason we need the DataValues dep?

mkborregaard:
No, I think it makes sense to remove it.

mkborregaard (approved):

